PR #24, opened 2020-04-17
channel-wise-ransac

### LINKED ISSUE #17: RANSAC consumes too much RAM
I have been testing pyprep against our datasets (GRUNECO, University of Antioquia, Colombia). For example I have 58 channels @ 1000Hz for about 5 minutes. This will need like 10 gb of RAM awarded from the OS to python. This is quite a lot for personal computers.

I have two possible proposals to solve this:
- Run RANSAC in a channel wise manner. 
- Automatically identify the resample needed for the data to fit memory 

I like more the first proposal and @sappelhoff  agrees with this approach. This was discussed in #9 .

Once #9 detrend_correction is merged into master I will upload this code so it may be reviewed and tested. It considerably changes find_bad_by_ransac() and a tiny change in run_ransac()

yjmantilla:
Here is the changes I did to implement the channel-wise ransac. Thus would close #17 

The old function of ransac was renamed to the same name but with '_hungry' at the end.

I cant right now offer much explanation regarding why I changed what I changed but for now I can say that a channel correlation depends only on that channel prediction by ransac, so there is no need to do all the predictions at once (or at least this is the logic i remember thinking).

This pull request is not intended to be merged right away but to communicate the changes done so should be marked as a work-in-progress). Also we need testing of:
- speed (is it much slower?)
- does it gives the same results?
- what to do with the seed of the random generator (see #21 ) , specifically matlab uses 
```
stream = RandStream('mt19937ar', 'Seed', 435656);`
```
- Testing against autoreject's ransac with sphinx examples (I dont really know how this hole sphinx stuff work) (also in #21 )


sappelhoff:
> The old function of ransac was renamed to the same name but with '_hungry' at the end.

We don't need to keep the old function, if the new one improves the functionality --> that's why we use GIT :-) if we ever need the old one back, we can just go back in time and get it.

> (I dont really know how this hole sphinx stuff work)

you simply write a .py script, put it into the "examples" folder, and sphinx (sphinx-gallery actually) will run this script and render it as a documentation page. See this file for example: https://github.com/sappelhoff/pyprep/blob/master/examples/use_noisy_module.py

The only "new" thing to learn is how to write RST (see https://thomas-cokelaer.info/tutorials/sphinx/rest_syntax.html) .. but you can take most of what you need from the example file I linked :-)

> I cant right now offer much explanation regarding why I changed what I changed but for now I can say 

okay, I'll wait with a review until everything is a bit more clear. Let me know if you need help

yjmantilla:
> We don't need to keep the old function, if the new one improves the functionality --> that's why we use GIT :-) if we ever need the old one back, we can just go back in time and get it.

Yes but I did that because I currently dont know :-/ how to test the pipeline with both versions without just renaming the old one. 

I mean is there a way to do a test (measuring results and timings) that calls different versions of the same module , in this case pyprep with old ransac vs pyprep with new ransac.

Now that I think, this would be like designing a test only for the ransac that measures results and time and execute it from the master (old ransac) and again but from the new branch and compare outside of the test script. So am I right with this approach?

sappelhoff:
> Now that I think, this would be like designing a test only for the ransac that measures results and time. Then execute it from the master (old ransac) and again but from the new branch and compare outside of the test script. So am I right with this approach?

yes, that sounds like a good approach 👍

sappelhoff:
@yjmantilla would it perhaps be possible for you to move your "new changes" to a new PR? This PR contains many "old" changes, and apparently the rebasing didn't work (so it's currently showing me 12 changed files)

yjmantilla:
@sappelhoff  So Ive checked a bit the timings of the ransac done channel-wise and it takes about 5 to 6 times slower. Nevertheless it consumes much less ram which is important for long and highly sampled eeg registers.

I'm thinking that the ransac should be done in your way by default and if it gives a memory error then it should attempt to do it channel wise. I have not yet implemented that because there is a test fail in test_prep_pipeline. Apparently it is attempting to substract a 64 channel register to a 63 channel register. Now the 63 channel register apparently should be from the edf loaded but this doesn't make sense. When i do the test locally it correctly loads a 64 channel edf so im baffled by this error. 

Remote test:
```
raw = <RawEDF | S004R01.edf, 63 x 9760 (61.0 s), ~4.9 MB, data loaded>
```

Local test:
```
<RawEDF | S004R01.edf, 64 x 9760 (61.0 s), ~4.9 MB, data loaded>
```

sappelhoff:
Yes, that's one reason why I would recommend that you pick all your changes and start off from a completely new PR that is up to date with upstream master.

I think I fixed these 63 vs 64 issues in this PR: https://github.com/sappelhoff/pyprep/pull/31

yjmantilla:
Thanks, I thought I did the rebase successfully but it seems like I didnt. Will try that.

PR #25, opened 2020-04-27
Ransac test

yjmantilla:
Hi. Been really busy. This PR is for the RANSAC test that I will use to compare the old and the new versions of RANSAC and in the future the autoreject version too.

I included it in the examples as per the suggestion of using sphinx libraries but if it should go in the test section do tell me so.

Basically I just: 
- generate a ton of sine waves for the "good channels" and a subset of "bad channels" that are sawtooths. 
- Embed this in a montage so that the ransac is able to interpolate. 
- Apply the ransac and assert that the results are what I simulated.

I timed the RANSAC in the basic way using the time library. I'm aware that one could use more complex profilers and so on but I don't really have experience using those methods.

I applied black formatting so thats why other files are changed. The only file that I actually worked on is run_ransac.py in examples . I did a rebase before so all else should be in order.

So looking forward to your feedback.


examples/run_ransac.py
```diff
@@ -0,0 +1,85 @@
...
+# Authors: Yorguin Mantilla <yjmantilla@gmail.com>
+#
+# License: MIT
+# Based On: use_noisy_module.py
+
+###############################################################################
+# First we import what we need for this example.
+import numpy as np
+import mne
+from scipy import signal as sx
```
sappelhoff, line 19:
```suggestion
from scipy import signal as signal```

File: examples/run_ransac.py [hunk 2]
```diff
@@ -0,0 +1,85 @@
...
+import time as clk
```
sappelhoff, line 20:
```suggestion
import time
```

please avoid using custom "import names". We can and should use the common ones such as numpy as np, pandas as pd, etc.

and I am fine with "scipy.signal as signal" --> but as soon as we get too funky with these, it will negatively impact readability. E.g., every pythonista knows `time.time()` ... but when they see `clk.time()` they cannot be sure whether this is some custom func, or actually `time.time()` (they first have to look at imports again)

examples/run_ransac.py
```diff
@@ -0,0 +1,85 @@
...
+
+from pyprep.find_noisy_channels import NoisyChannels
+
+###############################################################################
+# Now let's make some arbitrary MNE raw object for demonstration purposes.
+# We will think of good channels as sine waves and bad channels correlated with
+# each other as sawtooths. The idea is that the RANSAC is able to identify
+# these channels. We will need to set a montage because the RANSAC needs to
+# interpolate.
```
sappelhoff, line 29: I suggest to rephrase using these two snippets:

We will simulate good channels as sine waves and will inject a few "bad" channels with sawtooth shape.

RANSAC will then identify the "bad" channels, because they will not be predicted well by a random subset of "good" channels.

examples/run_ransac.py
```diff
@@ -0,0 +1,85 @@
...
+
+sfreq = 1000.0
+
+montage = mne.channels.make_standard_montage("standard_1020")
```
sappelhoff, line 33:
```suggestion
# We need a montage, because RANSAC uses spherical splines for interpolation
montage = mne.channels.make_standard_montage("standard_1020")
```

examples/run_ransac.py
```diff
@@ -0,0 +1,85 @@
...
+
+ch_names = montage.ch_names
+
+n_chans = len(ch_names)
+
+info = mne.create_info(ch_names=ch_names, sfreq=sfreq, ch_types=["eeg"] * n_chans)
+
+time = np.arange(0, 60, 1.0 / sfreq)  # 60 seconds of recording
+n_bad_chans = 3  # num_good_chans = n_chans - num_bad_chans
```
sappelhoff, line 42:
```suggestion
n_bad_chans = 3
```

examples/run_ransac.py
```diff
@@ -0,0 +1,85 @@
...
+
+bad_channels = np.random.choice(np.arange(n_chans), n_bad_chans, replace=False)
+bad_channels = [int(i) for i in bad_channels]
```
sappelhoff, line 45: it doesn't seem like this line is needed, the bad_channels array should already be int

examples/run_ransac.py
```diff
@@ -0,0 +1,85 @@
...
+bad_ch_names = [ch_names[i] for i in bad_channels]
+freq_good = 20
+freq_bad = 20
+
+X = [
+    sx.sawtooth(2 * np.pi * freq_bad * time)
+    if i in bad_channels
+    else np.sin(2 * np.pi * freq_good * time)
+    for i in range(n_chans)
+]
```
sappelhoff, line 55:
```suggestion
# The frequency components to use in the signal for good and bad channels
freq_good = 20
freq_bad = 20

# Generate the data
X = [
    signal.sawtooth(2 * np.pi * freq_bad * time)
    if i in bad_channels
    else np.sin(2 * np.pi * freq_good * time)
    for i in range(n_chans)
]
```

examples/run_ransac.py
```diff
@@ -0,0 +1,85 @@
...
+X = 2e-5 * np.array(X) + 1e-5 * np.random.random(
+    (n_chans, time.shape[0])
+)  # Scale to EEG and add noise.
```
sappelhoff, line 58:
```suggestion
# Scale the signal amplitude and add noise.
X = 2e-5 * np.array(X) + 1e-5 * np.random.random(
    (n_chans, time.shape[0])
)  
```

examples/run_ransac.py
```diff
@@ -0,0 +1,85 @@
...
+###############################################################################
+# Assign the mne object to the :class:`NoisyChannels` class. The resulting object
+# will be the place where all following methods are performed.
+
+nd = NoisyChannels(raw)
+
+
+###############################################################################
+# Find all bad channels and print a summary
+start_time = clk.time()
```
sappelhoff, line 74:
```suggestion
start_time = time.time()
```

examples/run_ransac.py
```diff
@@ -0,0 +1,85 @@
...
+nd.find_bad_by_ransac()
+print("--- %s seconds ---" % (clk.time() - start_time))
```
sappelhoff, line 76:
```suggestion
print("--- %s seconds ---" % (time.time() - start_time))
```

sappelhoff:
I am merging this now and will fix it subsequently. Thanks @yjmantilla

PR #29, opened 2020-06-07
skip filter feature

yjmantilla:
Explanation:

If the multitaper filter is skipped there is no need to remove the trend at the beginning since the trend at that point is specifically removed for the multitaper filter and then added back. So if we skip the filter we can jump directly to the robust referencing.

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -2,4 +2,4 @@ Aamna Lawrence <aamna.lawrence@gmail.com>
 Adam Li <adam2392@gmail.com>
 Victor Xiang <xiangliang3151@gmail.com>
 Victor Xiang <xiangliang3151@gmail.com> <32950290+Nick3151@users.noreply.github.com>
-Yorguin Mantilla <36543115+yjmantilla@users.noreply.github.com>
+Yorguin Mantilla <yjmantilla@gmail.com>
--- a/.mailmap
+++ b/.mailmap
@@ -3,3 +3,4 @@ Adam Li <adam2392@gmail.com>
 Victor Xiang <xiangliang3151@gmail.com>
 Victor Xiang <xiangliang3151@gmail.com> <32950290+Nick3151@users.noreply.github.com>
 Yorguin Mantilla <yjmantilla@gmail.com>
+Yorguin Mantilla <yjmantilla@gmail.com> <36543115+yjmantilla@users.noreply.github.com>
--- a/docs/whats_new.rst
+++ b/docs/whats_new.rst
@@ -31,7 +31,7 @@ Current
 Changelog
 ~~~~~~~~~
 
-- It's now possible to pass an empty list to skip the line noise removal by `Yorguin Mantilla`_ (`#29 <https://github.com/sappelhoff/pyprep/pull/29>`_)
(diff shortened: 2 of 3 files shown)


yjmantilla:
This is related to #28

sappelhoff:
Cool, thanks for being so quick!

Can you add to the docstring for this method? Specifically, this part should be extended:

```
        - line_freqs : 1d array
            - A list of line frequencies to be removed

```

- `1d array` should become `array_like` (see [docs here](https://numpydoc.readthedocs.io/en/latest/format.html#id16))
- `A list of line frequencies to be removed` should become something like `list of floats indicating frequencies to be remove. Can be an empty list to skip this step.`

sappelhoff:
Oh and you could add an entry into "whats new" that it's now possible to pass an empty list to skip the line noise removal. It'd go under the "changelog" header, because it's a new feature (although a small one) :)

yjmantilla:
@sappelhoff  Done, awaiting further instructions.

.mailmap
```diff
@@ -2,4 +2,4 @@ Aamna Lawrence <aamna.lawrence@gmail.com>
 Adam Li <adam2392@gmail.com>
 Victor Xiang <xiangliang3151@gmail.com>
 Victor Xiang <xiangliang3151@gmail.com> <32950290+Nick3151@users.noreply.github.com>
-Yorguin Mantilla <36543115+yjmantilla@users.noreply.github.com>
+Yorguin Mantilla <yjmantilla@gmail.com>
```
sappelhoff, line 5: The `.mailmap` file exists to keep the git history clean, specifically `git shortlog --summary` --> for example when a single author makes commits from different computers or accounts, these commits may show up as if from a different person 

`.mailmap` can be used to merge these "identities" back to how they are supposed to be represented.

In your case, `git shortlog --summary` yields your full name ... and based on what you added to whats_new.rst a while back, I inferred that you want to be represented with just those two names.

Next to your name you can also determine the email address that you want to be the "representation" ... From your edit I assume you want that to be Gmail. To get that right however, you need to modify the syntax a bit.

See this example:

 https://github.com/mne-tools/mne-bids/blob/2f8ef1b7a9830f663c7ae7b5c7b6dc5a54f230e4/.mailmap#L13-L14

So this would be correct:
```suggestion
Yorguin Mantilla <yjmantilla@gmail.com>
Yorguin Mantilla <yjmantilla@gmail.com> <36543115+yjmantilla@users.noreply.github.com>
```

docs/whats_new.rst
```diff
@@ -31,7 +31,7 @@ Current
 Changelog
 ~~~~~~~~~
 
-- nothing yet
+- It's now possible to pass an empty list to skip the line noise removal by `Yorguin Mantilla`_ (`#29 <https://github.com/sappelhoff/pyprep/pull/29>`_)
```
sappelhoff, line 34:
```suggestion
- It's now possible to pass an empty list for the ``line_freqs`` param in ``PrepPipeline`` to skip the line noise removal, by `Yorguin Mantilla`_ (`#29 <https://github.com/sappelhoff/pyprep/pull/29>`_)
```

sappelhoff:
@yjmantilla I just committed my remaining suggestions and merged. thanks for the PR, you can read what I said at a later point when you have time again :-)

yjmantilla:
Thank you. Indeed I have been really busy. I will be more mindful then to give a seed to the random generators of those tests. Makes a lot of sense.

PR #31, opened 2020-06-18
Give each class a random_state param to pass around to methods

sappelhoff:
This should fix our random tests to a certain seed so that they consistently pass according to our expected outputs.


sappelhoff:
@yjmantilla this PR should solve the randomness in the tests ...

The idea is that you can pass a `random_seed` to any of the main classes `PrepPipeline`, `NoisyChannels`, or `Reference` and if that class internally initializes another class, or calls a method, ... the random_seed is passed on, and so the outputs stay reproducible according to the seed

PR #32, opened 2020-07-05
lesser ram ransac

### LINKED ISSUE #17: RANSAC consumes too much RAM

### LINKED ISSUE #19: Slight inconsistency in perform_reference
In reference.py we have the method perform_reference() which states the following:
```
        # Phase 1: Estimate the true signal mean with robust referencing
        self.robust_reference()
        if self.noisy_channels["bad_all"]:
            self.raw.info["bads"] = self.noisy_channels["bad_all"]
            self.raw.interpolate_bads() <----------------------------------NOTICE THIS
        self.reference_signal = (
            np.nanmean(self.raw.get_data(picks=self.reference_channels), axis=0) * 1e6
        )
        rereferenced_index = [
            self.ch_names_eeg.index(ch) for ch in self.rereferenced_channels
        ]
        self.EEG = self.remove_reference(
            self.EEG, self.reference_signal, rereferenced_index
        )
```

Now in the original Matlab code this is equivalent to doRobustPost() inside performReference.m:

```
    function [] = doRobustPost()
        % Robust reference with interpolation afterwards
        referenceOut = robustReference(signal, referenceOut);
        noisy = referenceOut.badChannels.all;
        if isempty(noisy)   %No noisy channels -- ordinary ref
            referenceOut.referenceSignal = ...
                nanmean(signal.data(referenceOut.referenceChannels, :), 1);
        else
            sourceChannels = setdiff(referenceOut.evaluationChannels, noisy);
            signalNew = interpolateChannels(signal, noisy, sourceChannels); <----- NOTICE THIS
            referenceOut.referenceSignal = ...
                nanmean(signalNew.data(referenceOut.referenceChannels, :), 1);
            clear signalNew;
        end
        signal = removeReference(signal, referenceOut.referenceSignal, ...
            referenceOut.rereferencedChannels);
        referenceOut.noisyStatistics  = ...
            findNoisyChannels(removeTrend(signal, referenceOut), referenceOut);
        
```

So the basic difference is that robustReference produce a list of bad channels and this are interpolated in a dummy variable then averaged to produce a robust reference. In pyprep the interpolation is done on the raw object directly, not on a dummy copy of it.

I'm not so sure if this affects greatly or not but I would like your opinion on this.

If we wanted to make the code more alike the original one we would:

```
        # Phase 1: Estimate the true signal mean with robust referencing
        self.robust_reference()
        if self.noisy_channels["bad_all"]:
            dummy = self.raw.copy()
            dummy.info["bads"] = self.noisy_channels["bad_all"]
            dummy.interpolate_bads()
            self.reference_signal = (
                np.nanmean(dummy.get_data(picks=self.reference_channels), axis=0) * 1e6
            )
            del dummy
        else:
            self.reference_signal = (
                np.nanmean(self.raw.get_data(picks=self.reference_channels), axis=0)
                * 1e6
            )
        rereferenced_index = [
            self.ch_names_eeg.index(ch) for ch in self.rereferenced_channels
        ]
        self.EEG = self.remove_reference(
            self.EEG, self.reference_signal, rereferenced_index
        )

```

### LINKED ISSUE #28: Unable to finish running PyPREP
Whenever I try to run PyPREP I get the following output which seems promising, however it stalls here and continues to act as if it is running and no other output comes out. Any assistance would be greatly appreciated. (Also I tried running this on both the present master version v20, and also the development version and the same results occur in each one.)

EDIT by @sappelhoff : 2020-06-06 --> I added backticks ````` to format the following as a codeblock. Please see the [guidelines](https://help.github.com/en/github/writing-on-github/basic-writing-and-formatting-syntax#quoting-code) here @barlehmann

```
Extracting EDF parameters from /content/drive/My Drive/Colab Notebooks/Jacob5MeO2min.edf...
EDF file detected
Setting channel info structure...
Creating raw.info structure...
Reading 0 ... 200499  =      0.000 ...   801.996 secs...
<Info | 7 non-empty values
 bads: []
 ch_names: fp1, fp2, f3, f4, c3, c4, p3, p4, o1, o2, f7, f8, t3, t4, t5, ...
 chs: 19 EEG
 custom_ref_applied: False
 highpass: 0.0 Hz
 lowpass: 128.0 Hz
 meas_date: unspecified
 nchan: 19
 projs: []
 sfreq: 256.0 Hz
>
[<DigPoint |        LPA : (-82.5, -0.0, 0.0) mm     : head frame>, <DigPoint |     Nasion : (0.0, 102.7, 0.0) mm      : head frame>, <DigPoint |        RPA : (82.2, 0.0, 0.0) mm       : head frame>, <DigPoint |     EEG #1 : (-28.2, 102.3, 31.7) mm   : head frame>, <DigPoint |     EEG #3 : (28.6, 103.2, 31.6) mm    : head frame>, <DigPoint |    EEG #16 : (-67.4, 62.3, 30.5) mm    : head frame>, <DigPoint |    EEG #18 : (-48.2, 76.3, 80.9) mm    : head frame>, <DigPoint |    EEG #20 : (0.3, 83.3, 103.8) mm     : head frame>, <DigPoint |    EEG #22 : (49.7, 77.4, 79.6) mm     : head frame>, <DigPoint |    EEG #24 : (70.0, 64.2, 29.8) mm     : head frame>, <DigPoint |    EEG #40 : (-62.7, 16.1, 106.8) mm   : head frame>, <DigPoint |    EEG #42 : (0.4, 21.0, 140.9) mm     : head frame>, <DigPoint |    EEG #44 : (64.3, 16.7, 106.0) mm    : head frame>, <DigPoint |    EEG #62 : (-50.8, -48.7, 103.6) mm  : head frame>, <DigPoint |    EEG #64 : (0.3, -49.0, 129.2) mm    : head frame>, <DigPoint |    EEG #66 : (53.3, -48.5, 104.2) mm   : head frame>, <DigPoint |    EEG #81 : (-28.2, -84.3, 61.0) mm   : head frame>, <DigPoint |    EEG #83 : (28.6, -84.0, 60.9) mm    : head frame>, <DigPoint |    EEG #87 : (-80.7, 6.6, 36.6) mm     : head frame>, <DigPoint |    EEG #88 : (-69.4, -47.8, 47.3) mm   : head frame>, <DigPoint |    EEG #89 : (81.5, 7.5, 36.5) mm      : head frame>, <DigPoint |    EEG #90 : (70.0, -47.4, 47.3) mm    : head frame>]
[ 0  1  2  3  4  5  6  7  8  9 10 11 12 13 14 15 16 17 18]
['fp1', 'fp2', 'f3', 'f4', 'c3', 'c4', 'p3', 'p4', 'o1', 'o2', 'f7', 'f8', 't3', 't4', 't5', 't6', 'fz', 'cz', 'pz']
Setting up high-pass filter at 1 Hz

FIR filter parameters
---------------------
Designing a one-pass, zero-phase, non-causal highpass filter:
- Windowed time-domain design (firwin) method
- Hamming window with 0.0194 passband ripple and 53 dB stopband attenuation
- Lower passband edge: 1.00
- Lower transition bandwidth: 1.00 Hz (-6 dB cutoff frequency: 0.50 Hz)
- Filter length: 845 samples (3.301 sec)

<ipython-input-15-246072ead167>:34: DeprecationWarning: Using ``raise_if_subset`` to ``set_montage``  is deprecated and ``set_dig`` will be  removed in 0.21
  raw.set_montage(montage, match_case=False, raise_if_subset=False)
Setting up high-pass filter at 1 Hz

FIR filter parameters
---------------------
Designing a one-pass, zero-phase, non-causal highpass filter:
- Windowed time-domain design (firwin) method
- Hamming window with 0.0194 passband ripple and 53 dB stopband attenuation
- Lower passband edge: 1.00
- Lower transition bandwidth: 1.00 Hz (-6 dB cutoff frequency: 0.50 Hz)
- Filter length: 845 samples (3.301 sec)
```

yjmantilla:
# PR Description

Adding a lesser hungry version of the ransac albeit slower since it is done channel by channel. As of now it will try to execute the fast ransac and if that fails will try to do the channel-wise ransac. It is about 5 to 6 times slower than the ram hungry version.

Would close #17 and close #19 also closes #28 

Edit: Optimized, now ransac will automatically find the largest chunk of channels it can take at a time without running out of ram.

### DIFF
--- a/pyprep/find_noisy_channels.py
+++ b/pyprep/find_noisy_channels.py
@@ -10,7 +10,7 @@ from statsmodels import robust
 
 from pyprep.removeTrend import removeTrend
 from pyprep.utils import filter_design
-from pyprep.utils import SplitList
+from pyprep.utils import split_list
 
 
 class NoisyChannels:
@@ -469,7 +469,7 @@ class NoisyChannels:
 
         while mem_error:
             try:
-                channel_chunks = SplitList(job, chunk_size)
+                channel_chunks = split_list(job, chunk_size)
                 total_chunks = len(channel_chunks)
                 print("Total # of chunks:", total_chunks)
                 current = 1
--- a/pyprep/find_noisy_channels.py
+++ b/pyprep/find_noisy_channels.py
@@ -499,7 +499,8 @@ class NoisyChannels:
                 )
                 if chunk_size == 0:
                     raise MemoryError(
-                        "Not even doing 1 channel at a time the data fits in ram...\nMaybe try lowering the sample frequency."
+                        "Not even doing 1 channel at a time the data fits in ram..."
+                        "\nMaybe try lowering the sample frequency."
                     )
 
         # Thresholding
--- a/pyprep/find_noisy_channels.py
+++ b/pyprep/find_noisy_channels.py
@@ -367,6 +367,8 @@ class NoisyChannels:
         set_correlation = set(self.bad_by_correlation)
         not_hf = set_correlation - set_hf
         self.bad_by_SNR = self.bad_by_hf_noise + list(not_hf)
+        # this could have been done by an intersection
+        # self.bad_by_SNR = set_correlation.intersection(set_hf)
         return None
 
     def find_bad_by_ransac(
@@ -439,105 +441,58 @@ class NoisyChannels:
                 " quality tests."
             )
 
+        # Correlation windows setup
+        correlation_frames = corr_window_secs * self.sample_rate
+        correlation_window = np.arange(correlation_frames)
+        n = correlation_window.shape[0]
+        correlation_offsets = np.arange(
+            0, (self.signal_len - correlation_frames), correlation_frames
+        )
+        w_correlation = correlation_offsets.shape[0]
+
+        # Preallocate
+        channel_correlations = np.ones((w_correlation, self.n_chans_new))
+        # Notice self.EEGData.shape[0] = self.n_chans_new
+        # They came from the same drop of channels
+
+        print("Executing RANSAC\nThis may take a while, so be patient...")
+
         try:
             if channel_wise:
                 print("Forcing channel-wise ransac.")
                 raise MemoryError("Forcing channel-wise ransac.")
-            # Make the ransac predictions
-            ransac_eeg = self.run_ransac(
-                chn_pos=chn_pos,
-                chn_pos_good=chn_pos_good,
-                good_chn_labs=good_chn_labs,
-                n_pred_chns=n_pred_chns,
-                data=self.EEGData,
-                n_samples=n_samples,
+
+            chans_to_predict = list(range(self.n_chans_new))
+            channel_correlations = self.ransac_correlations(
+                chans_to_predict,
+                chn_pos,
+                chn_pos_good,
+                good_chn_labs,
+                n_pred_chns,
+                self.EEGData,
+                n_samples,
+                n,
+                w_correlation,
             )
-
-            # Correlate ransac prediction and eeg data
-            correlation_frames = corr_window_secs * self.sample_rate
-            correlation_window = np.arange(correlation_frames)
-            n = correlation_window.shape[0]
-            correlation_offsets = np.arange(
-                0, (self.signal_len - correlation_frames), correlation_frames
-            )
-            w_correlation = correlation_offsets.shape[0]
-
-            # For the actual data
-            data_window = self.EEGData[: self.n_chans_new, : n * w_correlation]
-            data_window = data_window.reshape(self.n_chans_new, n, w_correlation)
-
-            # For the ransac predicted eeg
-            pred_window = ransac_eeg[: self.n_chans_new, : n * w_correlation]
-            pred_window = pred_window.reshape(self.n_chans_new, n, w_correlation)
-
-            # Preallocate
-            channel_correlations = np.ones((w_correlation, self.n_chans_new))
-
-            # Perform correlations
-            for k in range(w_correlation):
-                data_portion = data_window[:, :, k]
-                pred_portion = pred_window[:, :, k]
-
-                R = np.corrcoef(data_portion, pred_portion)
-
-                # Take only correlations of data with pred
-                # and use diag to exctract correlation of
-                # data_i with pred_i
-                R = np.diag(R[0 : self.n_chans_new, self.n_chans_new :])
-                channel_correlations[k, :] = R
 
         except MemoryError:
             print("Cannot allocate enough ram for optimized ransac.")
-            print("Attempting slow ransac.")
-            # Correlate ransac prediction and eeg data
-            correlation_frames = corr_window_secs * self.sample_rate
-            correlation_window = np.arange(correlation_frames)
-            n = correlation_window.shape[0]
-
-            correlation_offsets = np.arange(
-                0, (self.signal_len - correlation_frames), correlation_frames
-            )
-            w_correlation = correlation_offsets.shape[0]
-
-            # Preallocate
-            channel_correlations = np.ones((w_correlation, self.n_chans_new))
-
-            for chanx in range(self.EEGData.shape[0]):
-                # Make the ransac predictions
-                ransac_eeg = self.run_ransac(
-                    chn_pos=chn_pos[[chanx], :],
-                    chn_pos_good=chn_pos_good,
-                    good_chn_labs=good_chn_labs,
-                    n_pred_chns=n_pred_chns,
-                    data=self.EEGData,
-                    n_samples=n_samples,
+            print("Attempting channel-wise ransac... it is slower :( .")
+            print("\nRANSAC PREDICTIONS:" + str(self.n_chans_new))
+            for chanx in range(self.n_chans_new):
+                chans_to_predict = [chanx]
+                channel_correlations[:, chanx] = self.ransac_correlations(
+                    chans_to_predict,
+                    chn_pos,
+                    chn_pos_good,
+                    good_chn_labs,
+                    n_pred_chns,
+                    self.EEGData,
+                    n_samples,
+                    n,
+                    w_correlation,
                 )
-
-                # For the actual data
-                data_window = self.EEGData[chanx, : n * w_correlation]
-                data_window = data_window.reshape(1, n, w_correlation)
-
-                # For the ransac predicted eeg
-                pred_window = ransac_eeg[0, : n * w_correlation]
-                pred_window = pred_window.reshape(1, n, w_correlation)
-
-                # Perform correlations
-                for k in range(w_correlation):
-                    data_portion = data_window[:, :, k]
-                    pred_portion = pred_window[:, :, k]
-
-                    R = np.corrcoef(data_portion, pred_portion)
-                    # Take only correlations of data with pred
-                    # and use diag to exctract correlation of
-                    # data_i with pred_i
-
-                    # R in this case is a 2x2 matrix with main diagonal elements = 1
-                    # The counter diagonal elements are the values we are interested in
-                    R = np.diag(R[0:1, 1:])
-                    # This transformation get us to the diagonal elements
-                    # although we could do it by slicing too
-                    channel_correlations[k, chanx] = R
-
+                print(chanx, end=" ", flush=True)
         # Thresholding
         thresholded_correlations = channel_correlations < corr_thresh
         frac_bad_corr_windows = np.mean(thresholded_correlations, axis=0)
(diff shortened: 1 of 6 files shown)


sappelhoff:
Cool, thanks @yjmantilla let me know once you are done and I should review

yjmantilla:
Should be ready for review now. The basic idea is to do for each channel: 

- predict the channel using a subset of the others channels at certain windows
- correlate prediction and real data of the channel for each window
- fill channel_correlation matrix at the channel position for each window

Basically is just exploiting the fact that a column of the channel_correlations matrix depends only on the prediction done for a particular channel, not for all the other channels. So in this sense we predict each channel in different iterations instead of doing all of the predictions at once.

pyprep/find_noisy_channels.py
```diff
@@ -539,7 +599,10 @@ def run_ransac(
 
         # Memory seems to be fine ...
         # Make the predictions
-        n_chns, n_timepts = data.shape
+        # n_chns, n_timepts = data.shape
+        # 2 next lines should be equivalent but support single channel processing
+        n_timepts = data.shape[1]
+        n_chns = chn_pos.shape[0]
```
sappelhoff, line 605: why is `n_chns, n_timepts = data.shape` not working anymore for single chn proc?

pyprep/find_noisy_channels.py
```diff
@@ -525,7 +583,9 @@ def run_ransac(
         # Before running, make sure we have enough memory
         try:
             available_gb = virtual_memory().available * 1e-9
-            needed_gb = (data.nbytes * 1e-9) * n_samples
+            needed_gb = (
+                data[: chn_pos.shape[0], :].nbytes * 1e-9
+            ) * n_samples  # chn_pos.shape[0] is just n_chans
```
sappelhoff, line 588:
```suggestion
            needed_gb = (
                data[: n_chans, :].nbytes * 1e-9
            ) * n_samples
```

If you had to insert a comment like that, that's a "code smell". (Not your fault, you just put the finger on it)

I suggest to put a line above the "try:"

```Python
n_chans = chn_pos.shape[0]
```

We need `n_chans` further down below anyhow - so this is a better point to introduce it.

pyprep/find_noisy_channels.py
```diff
@@ -438,48 +439,104 @@ def find_bad_by_ransac(
...
+            # For the actual data
+            data_window = self.EEGData[: self.n_chans_new, : n * w_correlation]
+            data_window = data_window.reshape(self.n_chans_new, n, w_correlation)
+
+            # For the ransac predicted eeg
+            pred_window = ransac_eeg[: self.n_chans_new, : n * w_correlation]
+            pred_window = pred_window.reshape(self.n_chans_new, n, w_correlation)
+
+            # Preallocate
+            channel_correlations = np.ones((w_correlation, self.n_chans_new))
```
sappelhoff, line 474: In this `Try` part, there are many lines of code that are then repeated in the `Except` part. Code duplication is almost never a good idea --> if you need a piece of code several times, make a short (private) function out of it, and re-use it.

pyprep/find_noisy_channels.py
yjmantilla: Because n_chns corresponds to the channels being predicted, not all the set of channels. Before since we did all the predictions at once the numbers were the same but now we only predict one channel at a time. 

Now, the number of channels predicted is defined through the chn_pos we want to interpolate. Currently this will always yield 1 since is only 1 channel.

I decided to keep it general because I was thinking a more optimized way of doing this was to predict in bash of m channels that doesn't go above the ram supplied. But that may be in the future.

pyprep/find_noisy_channels.py
yjmantilla: yeah, I was thinking the same thing when I did it

pyprep/find_noisy_channels.py
yjmantilla: I will see how can generalize it, indeed is almost the same except for some crucial parameters.

pyprep/find_noisy_channels.py
yjmantilla: Actually I ended up doing the optimization. I think the solution is quite elegant. Looking forward to your suggestions.

pyprep/utils.py
```diff
@@ -64,3 +64,29 @@ def filter_design(N_order, amp, freq):
     )
     kernel = np.multiply(kernel[0 : N_order + 1], (np.transpose(hamming_window[:])))
     return kernel
+
+
+def SplitList(mylist, chunk_size):
```
sappelhoff, line 69: why call it in CamelCase with Capital first letter? according to [pep8](https://www.python.org/dev/peps/pep-0008/#naming-conventions), this should rather be `split_list`

pyprep/utils.py
```diff
@@ -64,3 +64,29 @@ def filter_design(N_order, amp, freq):
...
+    Returns
+    -------
+    chunked lists: list
+        list of the chunked lists.
+
+
+    Code taken from:
+    https://stackoverflow.com/questions/312443/how-do-you-split-a-list-into-evenly-sized-chunks
+
+    Authors: atzz,cardamom from stackoverflow.
```
sappelhoff, line 88:
```suggestion
```

pyprep/utils.py
```diff
@@ -64,3 +64,29 @@ def filter_design(N_order, amp, freq):
     )
     kernel = np.multiply(kernel[0 : N_order + 1], (np.transpose(hamming_window[:])))
     return kernel
+
+
+def SplitList(mylist, chunk_size):
+    """Split list in chunks.
+
+    Parameters
```
sappelhoff, line 72:
```suggestion
    """Split list in chunks.

    See: https://stackoverflow.com/a/312466/5201771

    Parameters
```

Small code snippets like this one dont have to be cited with name and everything. A link to the stackoverflow post should be enough.

You can generate a better link (like I did for you now) by clicking on "share" as in the left lower corner of this screenshot:

image

pyprep/find_noisy_channels.py
```diff
@@ -367,6 +370,8 @@ def find_bad_by_SNR(self):
         set_correlation = set(self.bad_by_correlation)
         not_hf = set_correlation - set_hf
         self.bad_by_SNR = self.bad_by_hf_noise + list(not_hf)
+        # this could have been done by an intersection
+        # self.bad_by_SNR = set_correlation.intersection(set_hf)
```
sappelhoff, line 374: feel free to improve :+1:

pyprep/find_noisy_channels.py
```diff
@@ -287,7 +288,9 @@ def find_bad_by_correlation(
         frac_bad : float
             percentage of data windows in which the correlation threshold was
             not surpassed and if a channel gets a value of greater than 1%, it
-            is designated bad.
+            is designated bad. Notice that if the # of windows is low the default
+            frac_bad may be too extreme (ie 60 windows imply only 1 bad window
+            needed).
```
sappelhoff, line 293:
```suggestion
            is designated bad. Notice that if `correlation_secs` is high, and
            thus the number of windows is low, the default `frac_bad` may
            be too extreme. For example if only 60 windows are available
            for a dataset, only a single bad window would be needed to
            classify as bad.
```

pyprep/utils.py
yjmantilla: explanation: just my laziness

yjmantilla:
I'm actually wondering now if it is faster to decrease the chunk size by 1 or to divide it by half. Apparently it is faster to find the largest possible chunk which may be missed when dividing by half.

sappelhoff:
> I'm actually wondering now if it is faster to decrease the chunk size by 1 or to divide it by half. Apparently it is faster to find the largest possible chunk which may be missed when dividing by half.

you made it work now --> so that's already a nice step. I think there would be other issues to focus on now, instead of optimizing this fix even more :-)

Are you ready with all changes now?

yjmantilla:
Almost ready, in the end I couldn't help but implement the optimization lol. Also made the output a bit less verbose and straight to the optimal value.

yjmantilla:
ok, it is ready if there are no further suggestions.

sappelhoff:
no more suggestions, thanks for the fix!

PR #34, opened 2020-08-04
Adding a full_raw property to PrepPipeline

### LINKED ISSUE #33: EOG get dropped
I suppose most people would want to use PyPrep as an early preprocessing pipeline to get ride of line noise and get a robust average reference (at least, that is my understanding from the documentation). To do that, it seems that PyPrep should make available a raw object that has been cleaned and re-referenced. My understanding is that it does, as the .raw attribute of the PrepPipeline class. However, in this Raw object, some channels (like EOG channels) are dropped by PyPrep. I don't think that they should be, since they are likely to be necessary for the following artifact rejection steps that will be run after PyPrep.

christian-oreilly:
Fix #33.

### DIFF
--- a/docs/whats_new.rst
+++ b/docs/whats_new.rst
@@ -34,7 +34,7 @@ Changelog
 ~~~~~~~~~
 - Adding ``ch_types_all`` and ``ch_names_non_eeg`` attributes to ``PrepPipeline``, by `Yorguin Mantilla`_ (`#34 <https://github.com/sappelhoff/pyprep/pull/34>`_)
 - Renaming of ``ch_names`` attribute of ``PrepPipeline`` to ``ch_names_all``, by `Yorguin Mantilla`_ (`#34 <https://github.com/sappelhoff/pyprep/pull/34>`_)
-- It's now possible to pass 'eeg' ``to ref_chs`` and ``reref_chs`` parameters to select only (all) eeg channels for referencing, by `Yorguin Mantilla`_ (`#34 <https://github.com/sappelhoff/pyprep/pull/34>`_)
+- It's now possible to pass ``'eeg'`` to ``ref_chs`` and ``reref_chs`` parameters to select only eeg channels for referencing, by `Yorguin Mantilla`_ (`#34 <https://github.com/sappelhoff/pyprep/pull/34>`_)
 - Prep will retain the non eeg channels through the ``raw`` attribute. The eeg-only and non-eeg parts will be in raw_eeg and raw_non_eeg respectively. See the ``raw`` attribute, by `Christian Oreilly`_ (`#34 <https://github.com/sappelhoff/pyprep/pull/34>`_)
 - When a ransac call needs more memory than available, pyprep will now automatically switch to a slower but less memory-consuming version of ransac. See the ``channel_wise``parameter, by `Yorguin Mantilla`_ (`#32 <https://github.com/sappelhoff/pyprep/pull/32>`_)
 - It's now possible to pass an empty list for the ``line_freqs`` param in ``PrepPipeline`` to skip the line noise removal, by `Yorguin Mantilla`_ (`#29 <https://github.com/sappelhoff/pyprep/pull/29>`_)
--- a/docs/whats_new.rst
+++ b/docs/whats_new.rst
@@ -35,7 +35,7 @@ Changelog
 - Adding ``chs_types`` and ``chs_non_eeg`` attributes to ``PrepPipeline``, by `Yorguin Mantilla`_ (`#34 <https://github.com/sappelhoff/pyprep/pull/34>`_)
 - Renaming of ``ch_names`` and ``ch_names_eeg`` attributes of ``PrepPipeline`` to ``chs_all`` and ``chs_eeg``, by `Yorguin Mantilla`_ (`#34 <https://github.com/sappelhoff/pyprep/pull/34>`_)
 - It's now possible to pass 'eeg' ``to ref_chs`` and ``reref_chs`` parameters to select only (all) eeg channels for referencing, by `Yorguin Mantilla`_ (`#34 <https://github.com/sappelhoff/pyprep/pull/34>`_)
-- Prep will retain the non eeg channels through the attribute full_raw. See the ``full_raw``attribute, by `Christian Oreilly`_ (`#34 <https://github.com/sappelhoff/pyprep/pull/34>`_)
+- Prep will retain the non eeg channels through the attribute full_raw. See the ``full_raw`` attribute, by `Christian Oreilly`_ (`#34 <https://github.com/sappelhoff/pyprep/pull/34>`_)
 - When a ransac call needs more memory than available, pyprep will now automatically switch to a slower but less memory-consuming version of ransac. See the ``channel_wise``parameter, by `Yorguin Mantilla`_ (`#32 <https://github.com/sappelhoff/pyprep/pull/32>`_)
 - It's now possible to pass an empty list for the ``line_freqs`` param in ``PrepPipeline`` to skip the line noise removal, by `Yorguin Mantilla`_ (`#29 <https://github.com/sappelhoff/pyprep/pull/29>`_)
 - The three main classes ``PrepPipeline``, ``NoisyChannels``, and ``Reference`` now have a ``random_state`` parameter to set a seed that gets passed on to all their internal methods and class calls, by `Stefan Appelhoff`_ (`#31 <https://github.com/sappelhoff/pyprep/pull/31>`_)
--- a/docs/whats_new.rst
+++ b/docs/whats_new.rst
@@ -35,7 +35,7 @@ Changelog
 - Adding ``ch_types_all`` and ``ch_names_non_eeg`` attributes to ``PrepPipeline``, by `Yorguin Mantilla`_ (`#34 <https://github.com/sappelhoff/pyprep/pull/34>`_)
 - Renaming of ``ch_names`` attribute of ``PrepPipeline`` to ``ch_names_all``, by `Yorguin Mantilla`_ (`#34 <https://github.com/sappelhoff/pyprep/pull/34>`_)
 - It's now possible to pass 'eeg' ``to ref_chs`` and ``reref_chs`` parameters to select only (all) eeg channels for referencing, by `Yorguin Mantilla`_ (`#34 <https://github.com/sappelhoff/pyprep/pull/34>`_)
-- Prep will retain the non eeg channels through the attribute full_raw. See the ``full_raw`` attribute, by `Christian Oreilly`_ (`#34 <https://github.com/sappelhoff/pyprep/pull/34>`_)
+- Prep will retain the non eeg channels through the ``raw`` attribute. The eeg-only and non-eeg parts will be in raw_eeg and raw_non_eeg respectively. See the ``raw`` attribute, by `Christian Oreilly`_ (`#34 <https://github.com/sappelhoff/pyprep/pull/34>`_)
 - When a ransac call needs more memory than available, pyprep will now automatically switch to a slower but less memory-consuming version of ransac. See the ``channel_wise``parameter, by `Yorguin Mantilla`_ (`#32 <https://github.com/sappelhoff/pyprep/pull/32>`_)
 - It's now possible to pass an empty list for the ``line_freqs`` param in ``PrepPipeline`` to skip the line noise removal, by `Yorguin Mantilla`_ (`#29 <https://github.com/sappelhoff/pyprep/pull/29>`_)
 - The three main classes ``PrepPipeline``, ``NoisyChannels``, and ``Reference`` now have a ``random_state`` parameter to set a seed that gets passed on to all their internal methods and class calls, by `Stefan Appelhoff`_ (`#31 <https://github.com/sappelhoff/pyprep/pull/31>`_)
--- a/docs/whats_new.rst
+++ b/docs/whats_new.rst
@@ -19,9 +19,11 @@ People who contributed to this software across releases (in alphabetical order)
 
 * `Aamna Lawrence`_
 * `Adam Li`_
+* `Christian Oreilly`_
 * `Stefan Appelhoff`_
 * `Victor Xiang`_
 * `Yorguin Mantilla`_
+
 
 .. _current:
 
@@ -30,6 +32,10 @@ Current
 
 Changelog
 ~~~~~~~~~
+- Adding ``chs_types`` and ``chs_non_eeg`` attributes to ``PrepPipeline``, by `Yorguin Mantilla`_ (`#34 <https://github.com/sappelhoff/pyprep/pull/34>`_)
+- Renaming of ``ch_names`` and ``ch_names_eeg`` attributes of ``PrepPipeline`` to ``chs_all`` and ``chs_eeg``, by `Yorguin Mantilla`_ (`#34 <https://github.com/sappelhoff/pyprep/pull/34>`_)
+- It's now possible to pass 'eeg' ``to ref_chs`` and ``reref_chs`` parameters to select only (all) eeg channels for referencing, by `Yorguin Mantilla`_ (`#34 <https://github.com/sappelhoff/pyprep/pull/34>`_)
+- Prep will retain the non eeg channels through the attribute full_raw. See the ``full_raw``attribute, by `Christian Oreilly`_ (`#34 <https://github.com/sappelhoff/pyprep/pull/34>`_)
 - When a ransac call needs more memory than available, pyprep will now automatically switch to a slower but less memory-consuming version of ransac. See the ``channel_wise``parameter, by `Yorguin Mantilla`_ (`#32 <https://github.com/sappelhoff/pyprep/pull/32>`_)
 - It's now possible to pass an empty list for the ``line_freqs`` param in ``PrepPipeline`` to skip the line noise removal, by `Yorguin Mantilla`_ (`#29 <https://github.com/sappelhoff/pyprep/pull/29>`_)
 - The three main classes ``PrepPipeline``, ``NoisyChannels``, and ``Reference`` now have a ``random_state`` parameter to set a seed that gets passed on to all their internal methods and class calls, by `Stefan Appelhoff`_ (`#31 <https://github.com/sappelhoff/pyprep/pull/31>`_)
@@ -108,5 +114,6 @@ Changelog
 .. _Stefan Appelhoff: http://stefanappelhoff.com/
 .. _Aamna Lawrence: https://github.com/AamnaLawrence
 .. _Adam Li: https://github.com/adam2392/
+.. _Christian Oreilly: https://github.com/christian-oreilly
 .. _Victor Xiang: https://github.com/Nick3151
 .. _Yorguin Mantilla: https://github.com/yjmantilla
--- a/docs/whats_new.rst
+++ b/docs/whats_new.rst
@@ -32,7 +32,7 @@ Current
 
 Changelog
 ~~~~~~~~~
-- Adding ``ch_types_all`` and ``ch_names_non_eeg`` attributes to ``PrepPipeline``, by `Yorguin Mantilla`_ (`#34 <https://github.com/sappelhoff/pyprep/pull/34>`_)
+- Channel types  are now available from a new ``ch_types_all`` attribute, and non-EEG channel names are now available from a new ``ch_names_non_eeg`` attribute from :class:`pyprep.PrepPipeline`, by `Yorguin Mantilla`_ (`#34 <https://github.com/sappelhoff/pyprep/pull/34>`_)
 - Renaming of ``ch_names`` attribute of ``PrepPipeline`` to ``ch_names_all``, by `Yorguin Mantilla`_ (`#34 <https://github.com/sappelhoff/pyprep/pull/34>`_)
 - It's now possible to pass ``'eeg'`` to ``ref_chs`` and ``reref_chs`` parameters to select only eeg channels for referencing, by `Yorguin Mantilla`_ (`#34 <https://github.com/sappelhoff/pyprep/pull/34>`_)
 - Prep will retain the non eeg channels through the ``raw`` attribute. The eeg-only and non-eeg parts will be in raw_eeg and raw_non_eeg respectively. See the ``raw`` attribute, by `Christian Oreilly`_ (`#34 <https://github.com/sappelhoff/pyprep/pull/34>`_)
--- a/docs/whats_new.rst
+++ b/docs/whats_new.rst
@@ -32,8 +32,8 @@ Current
 
 Changelog
 ~~~~~~~~~
-- Adding ``chs_types`` and ``chs_non_eeg`` attributes to ``PrepPipeline``, by `Yorguin Mantilla`_ (`#34 <https://github.com/sappelhoff/pyprep/pull/34>`_)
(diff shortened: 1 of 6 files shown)


pyprep/prep_pipeline.py
```diff
@@ -56,14 +56,32 @@ def __init__(self, raw, prep_params, montage, ransac=True, random_state=None):
         self.raw = raw.copy()
         self.ch_names = self.raw.ch_names
         self.raw.set_montage(montage)
-        self.raw.pick_types(eeg=True, eog=False, meg=False)
+        self.raw_non_eeg = self.raw.copy()
```
sappelhoff, line 59: is it necessary to make a full copy here? given that we (used to?) run into RAM issues in pyprep regularly, I'd prefer to not copy if possible.

E.g., *extract* the non-EEG channels without a copy and later add them.

pyprep/prep_pipeline.py
```diff
@@ -56,14 +56,32 @@ def __init__(self, raw, prep_params, montage, ransac=True, random_state=None):
...
+
+        ch_types = ["eeg", "meg", "stim", "eog", "ecg", "emg", "ref_meg",
+                    "misc", "resp", "syst", "seeg", "dipole", "gof",
+                    "bio", "ecog", "fnirs", "csd"]
```
sappelhoff, line 63: given that we work reasonably close with MNE, I wonder if it'd be good or bad to use private variable from MNE such as this one:

https://github.com/mne-tools/mne-python/blob/76fc838e954e70697f33cab0edbca61872ca1c23/mne/io/pick.py#L70

@larsoner do you perhaps have advice on this? We want a list of all channel types available in MNE.

larsoner: Perhaps the keys of the dict returned by `get_channel_type_constants()`

https://github.com/mne-tools/mne-python/blob/master/mne/io/pick.py#L18

yjmantilla:
@christian-oreilly if you want I could try to implement the changes needed for the tests to pass and the test of the eog not being dropped.

pyprep/prep_pipeline.py
yjmantilla: Problem would be that there are little discrepancies between the two:

the dict keys would be:
```
['bio', 'chpi', 'csd', 'dipole', 'ecg', 'ecog', 'eeg', 'emg', 'eog', 'exci', 'fnirs_od', 'fnirs_raw', 'gof', 'grad', 'hbo', 'hbr', 'ias', 'mag', 'misc', 'ref_meg', 'resp', 'seeg', 'stim', 'syst']
```
whereas the pick_types channel arguments are:
```
['bio', 'chpi', 'csd', 'dipole', 'ecg', 'ecog', 'eeg', 'emg', 'eog', 'exci', 'fnirs', 'gof', 'ias', 'meg', 'misc', 'ref_meg', 'resp', 'seeg', 'stim', 'syst']
```

So we would need little non-homogeneous corrections like:
- fnirs_od fnirs_raw to single fnirs
- delete grad hbo hbr
- replace mag with meg

Im not sure if this simplifies or complicates the stuff.

Another way is using inspect module:
```
ch_types = inspect.signature(mne.io.pick.pick_types)
ch_types = ch_types.parameters.items()
ch_types = list(ch_types)
ch_types = [x[0] for x in ch_types]
ch_types.pop(0) #remove info
ch_types = ch_types[:-3] #remove non channel parameters
```
Which Im also not sure if it is better than making the list like the initial approach

pyprep/prep_pipeline.py
yjmantilla: What do you mean "extract non-EEG channels without a copy" . Like only make a copy of the channel data rather than the raw object? Besides that I'm not sure how could we extract them.

Although there where many ram issues I think copying the raw object entirely does not compare to them. Usually the ram issues we had were in the other of 9gb or more (though they depend on the user ram capability). Moreover the raw_non_eeg copy a few lines later drops the eeg channels so the consumption gets lower. 

It would be possible to retain the channels without any copy by really verifying how the deeper pyprep code treats the raw object since the to-be-processed channels are passed as a list to some functions but I'm not entirely sure all the code works correctly like that, so thats a risky bet.

Last solution I can come up is to save the non eeg channels on disk to save ram.

In any case I don't think the only-non-eeg copy will be a problem.

pyprep/prep_pipeline.py
yjmantilla: Regarding the _first_rule private variable I dont know how could we extract it but if we manage to do it then the only change would be to add 'csd' to it.

pyprep/prep_pipeline.py
sappelhoff: > Like only make a copy of the non-eeg channel data rather than the raw object?

yeah, not even making a copy - simply splitting the existing raw into two: EEG ... and non-EEG.

But you raise valid points, the copy here is probably not a big problem, especially because the EEG channels are then dropped a few lines below. (as you say)

pyprep/prep_pipeline.py
sappelhoff: > Which Im also not sure if it is better than making the list like the initial approach

we can also stay with a list of strings instead of importing that list directly from MNE-Python ... but I wanted to check whether there is an easy way to directly import this --> it would decrease maintenance burden. 

IF we hardcode the list here, we'll have to update it every time something about these channel types changes in MNE

sappelhoff:
> @christian-oreilly if you want I could try to implement the changes needed for the tests to pass and the test of the eog not being dropped.

@yjmantilla I gave you maintainer rights for this repository, so as long as christian has the "allow edits from maintainers" box checked, you should be able to work on this PR.

christian-oreilly:
@yjmantilla Sure, thanks. I won't have time to look into this any time soon, so please feel free to use I submitted and change it however you see fit. The "allow edits from maintainers" is checked.

pyprep/prep_pipeline.py
yjmantilla: how do you feel about the inspect module way? Although it is a bit ugly I find the following positive points:

- If mne changes then the pick_type method would have changed accordingly.
- I think that params positions will stay the same , that is , info im almost sure it will stay in front and the other non chan_type params will be the last three --> include, exclude , selection.

We could also extract them from raw.pick_types rather than io.pick , then we would get rid of the last 4.

Cons:
- ugly
- maybe unpythonic
- confusing for someone just reading the code ?
- breaks if non chan_type params are added or change position

yjmantilla: Maybe another way would be using the include exclude parameters of pick, assuming 'eeg' types are assigned correctly which the other method also needs

- obtain list of eeg chans --> non eeg channels = all - eeg chans
- include only non eeg_channels for copy/extraction

I think this may be the one. Currently coding the implementation.

pyprep/prep_pipeline.py
yjmantilla: Tests passed, looking forward for the feedback. I'm not sure how to do such splitting but if you got an idea of the implementation I can change it.

pyprep/prep_pipeline.py
```diff
@@ -54,16 +54,38 @@ class PrepPipeline:
     def __init__(self, raw, prep_params, montage, ransac=True, random_state=None):
         """Initialize PREP class."""
         self.raw = raw.copy()
-        self.ch_names = self.raw.ch_names
+
+        # separe eeg and non eeg channels
+        self.raw_non_eeg = self.raw.copy()
+        self.chs_all = self.raw.ch_names.copy()
+        self.chs_types = self.raw.get_channel_types()
```
sappelhoff, line 61: These are *all* types, so I would find `ch_types_all` more straight forward to understand

also, for all of these variables, I would find `ch_...` instead of `chs_...` better. Speaking aloud it would be "channels types" versus "channel types ... and the latter makes more sense IMHO.

Note: I make one suggestion here to clarify my point, but adhering to this point would mean many more small changes in code to adjust all the variable and attribute names.
```suggestion
        self.ch_types_all = self.raw.get_channel_types()
```

pyprep/prep_pipeline.py
```diff
@@ -54,16 +54,38 @@ class PrepPipeline:
     def __init__(self, raw, prep_params, montage, ransac=True, random_state=None):
         """Initialize PREP class."""
         self.raw = raw.copy()
-        self.ch_names = self.raw.ch_names
+
+        # separe eeg and non eeg channels
+        self.raw_non_eeg = self.raw.copy()
+        self.chs_all = self.raw.ch_names.copy()
```
sappelhoff, line 60: maybe `ch_names_all` ? then we know immediately it's a list of channel names

pyprep/prep_pipeline.py
```diff
@@ -54,16 +54,38 @@ class PrepPipeline:
...
+        self.chs_types = self.raw.get_channel_types()
+        self.chs_eeg = [
+            self.chs_all[i]
+            for i in range(len(self.chs_all))
+            if self.chs_types[i] == "eeg"
+        ]
+        self.chs_non_eeg = set(self.chs_all) - set(self.chs_eeg)
```
sappelhoff, line 67: this would become `self.ch_names_non_eeg`

pyprep/prep_pipeline.py
```diff
@@ -54,16 +54,38 @@ class PrepPipeline:
...
+        self.raw.pick_channels(self.chs_eeg)
```
sappelhoff, line 68: this makes me feel like `self.raw` should actually be `self.raw_eeg`

pyprep/prep_pipeline.py
```diff
@@ -54,16 +54,38 @@ class PrepPipeline:
...
         self.raw.set_montage(montage)
-        self.raw.pick_types(eeg=True, eog=False, meg=False)
-        self.ch_names_eeg = self.raw.ch_names
+
         self.EEG_raw = self.raw.get_data() * 1e6
         self.prep_params = prep_params
+        if self.prep_params["ref_chs"] == "eeg":
+            self.prep_params["ref_chs"] = self.chs_eeg
+        if self.prep_params["reref_chs"] == "eeg":
+            self.prep_params["reref_chs"] = self.chs_eeg
```
sappelhoff, line 78: This needs to be updated in the docstring above, and I suggest to make this the new default.

```Python
        - ref_chs : list | 'eeg'
            - A list of channel names to be used for rereferencing.
               Can be a str 'eeg' to use all EEG channels (default).
        - reref_chs : list | 'eeg'
            - A list of channel names to be used for line-noise removed, and
              referenced. Can be a str 'eeg' to use all EEG channels (default).
```

pyprep/prep_pipeline.py
```diff
@@ -54,16 +54,38 @@ class PrepPipeline:
...
         self.sfreq = self.raw.info["sfreq"]
         self.ransac = ransac
         self.random_state = check_random_state(random_state)
 
+    @property
+    def full_raw(self):
+        """Return a version of self.raw that includes the non-eeg channels."""
+        full_raw = self.raw.copy()
+        return full_raw.add_channels([self.raw_non_eeg])
```
sappelhoff, line 87: I think we should change the way we refer to the different `raw` objects.

IMHO we should have:
1. the raw object with only EEG: `raw_eeg`
1. with all non-eeg: `raw_non_eeg`
1. the full raw, which is here called `full_raw` ... I suggest we call this just `raw`

furthermore, this attribute needs to be documented (like all other publicly accessible attributes actually) in the Docstr of the class.

For an example, see here: https://github.com/mne-tools/mne-python/blob/8fd3c1096c07c1a6dfaa7bb2a4021ba97f02b4fb/mne/io/brainvision/brainvision.py#L54-L57

tests/test_prep_pipeline.py
```diff
@@ -204,3 +204,47 @@ def test_prep_pipeline(raw, montage):
...
+    # Numpy array of size 4 X 10000.
+    data = np.array([sin, cos, sinX2, cosX2])
+
+    # Definition of channel types and names.
+    ch_types = ["eog", "eog", "misc", "misc"]
+    ch_names_non_eeg = ["sin", "cos", "sinX2", "cosX2"]
+    info = mne.create_info(ch_names=ch_names_non_eeg, sfreq=sfreq, ch_types=ch_types)
+    raw_non_eeg = mne.io.RawArray(data, info)
+
+    raw_copy.add_channels([raw_non_eeg])
```
sappelhoff, line 230: Can you extend and re-use https://github.com/sappelhoff/pyprep/blob/6bb931c0c8c9739a1dd2b8e047995cb2387b3ede/tests/test_noisy.py#L12

for this?

docs/whats_new.rst
```diff
@@ -19,17 +19,23 @@ People who contributed to this software across releases (in alphabetical order)
...
 
+
 .. _current:
 
 Current
 -------
 
 Changelog
 ~~~~~~~~~
+- Adding ``chs_types`` and ``chs_non_eeg`` attributes to ``PrepPipeline``, by `Yorguin Mantilla`_ (`#34 <https://github.com/sappelhoff/pyprep/pull/34>`_)
```
sappelhoff, line 35: as also mentioned in a comment below: We should properly document all publicly accessible class attributes in the class docstr in an `Attributes` section.

docs/whats_new.rst
```diff
@@ -19,17 +19,23 @@ People who contributed to this software across releases (in alphabetical order)
...
+- Renaming of ``ch_names`` and ``ch_names_eeg`` attributes of ``PrepPipeline`` to ``chs_all`` and ``chs_eeg``, by `Yorguin Mantilla`_ (`#34 <https://github.com/sappelhoff/pyprep/pull/34>`_)
+- It's now possible to pass 'eeg' ``to ref_chs`` and ``reref_chs`` parameters to select only (all) eeg channels for referencing, by `Yorguin Mantilla`_ (`#34 <https://github.com/sappelhoff/pyprep/pull/34>`_)
```
sappelhoff, line 37: And I suggest to make this the new default

sappelhoff:
PS: Thanks for continuing this PR @yjmantilla :)

examples/run_full_prep.py
```diff
@@ -92,10 +92,11 @@
 
 # Fit prep
 prep_params = {
-    "ref_chs": ch_names_eeg,
-    "reref_chs": ch_names_eeg,
```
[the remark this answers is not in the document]
yjmantilla: ok, I will change it in the example but leave one with the 'eeg' and another with the channel list on the tests so that code coverage doesn't decline (as a result of the if statement selecting the case).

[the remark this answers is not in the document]
yjmantilla: Shouldn't we provide though the channel list way in an example too though?

pyprep/prep_pipeline.py
yjmantilla: I agree, to be honest I was sticking to the 'chs' way because the input parameters are 'ref_chs' and 'reref_chs' but in a way those are different grammar contexts lol

pyprep/prep_pipeline.py
yjmantilla: I was wondering the same thing. We should have a discussion on the naming of input output parameters. I remember one user confused becase the "raw" was actually the processed signal. Obviously from the developer pov we are referring to the raw mne object but the user is thinking of an unprocessed signal.

So, should we stick to the raw  prefix?

Now for the suffix: should the full_raw actually be raw_all?

pyprep/prep_pipeline.py
yjmantilla: I wanted to make it a default but we have the following problem, the input dictionary defines ref, reref and the line freqs. The ref and reref as we know are usually 'eeg' channels (unless is a really special case); but the line_freqs can be 50 or 60.., Which line_freq should we assume ? Would a default line_freq make errors more common as users can init the prep with no params?

yjmantilla: I will add the doc string without the default for now until we clarify this.

Also should we consider to pass them as normal function arguments rather than a dictionary to avoid the mentioned problem? We could even support both dictionary and argument inputs but maybe it would a little bit confusing. 

What do you think?

pyprep/prep_pipeline.py
yjmantilla: Will leave it as 'raw' until we decide the naming convention

pyprep/prep_pipeline.py
yjmantilla: lol, I read this after making my previous comments on the previous discussions. The way you are suggesting is consistent with what we currently have so I will make the change. But do take notice of the cognitive dissonance of 'raw' and a processed signal.

docs/whats_new.rst
yjmantilla: Should those attributes be public though, I mean should anyone actually use the channel names and types attributes?

tests/test_prep_pipeline.py
yjmantilla: I included it in the utils script. Tell me if the extension is what you thought. Had to input the time vector parameter to assure the exact same length of the time vector of the eeg signals.

sappelhoff:
I will review this next monday, sorry for the delay.

pyprep/utils.py
```diff
@@ -86,3 +87,63 @@ def split_list(mylist, chunk_size):
...
+    RNG: np.random.RandomState
+        Random State Seed
+    Returns
+    -------
+    raw : mne raw object
+        The mne object for performing the tests.
+
+    n_freq_comps : int
+
+    freq_range : list, len==2
```
sappelhoff, line 129:
```suggestion
    raw : mne raw object
        The mne object for performing the tests.
    n_freq_comps : int
        Number of random frequency components to introduce.
    freq_range : tuple
       The low (`freq_range[0]`) and high (`freq_range[1]`) endpoints of 
       a frequency range from which random draws will be made to 
       introduce frequency components in the test data.
```

pyprep/utils.py
```diff
@@ -86,3 +87,63 @@ def split_list(mylist, chunk_size):
...
+    sfreq : float
+        Sampling frequency associated with the time vector.
+    n_freq_comps : int
+        Number of signal components summed to make a signal.
+    freq_range : list, len==2
+        Signals will contain freqs from this range.
+    scale: float
+        scale of the signal in volts. (ie 1e-6 for microvolts).
+    RNG: np.random.RandomState
+        Random State Seed
```
sappelhoff, line 121:
```suggestion
    scale : float
        Scaling factor applied to the signal in volts. For example 1e-6 to
        get microvolts.
    RNG : np.random.RandomState
        Random state seed.
```

tests/test_noisy.py
```diff
@@ -5,84 +5,40 @@
...
+    "C5",
+    "C6",
+]
+ch_types = ["eeg" for chn in ch_names]
+
 while not found_good_test_object:
-    raw, n_freq_comps, freq_range = make_random_mne_object()
+    raw, n_freq_comps, freq_range = make_random_mne_object(
+        ch_names, ch_types, t, sfreq, RNG=RNG
+    )
```
sappelhoff, line 41: you introduced a random seed, so we can get rid of the while loop :)

just find a random seed that fulfills the condition, and set that random seed explicitly in the code --> then the while loop can be removed.

it was a very dirty coding practice to begin with, to be honest (my fault :laughing: )

pyprep/prep_pipeline.py
```diff
@@ -53,20 +83,50 @@ class PrepPipeline:
...
-        self.sfreq = self.raw.info["sfreq"]
+        if self.prep_params["ref_chs"] == "eeg":
+            self.prep_params["ref_chs"] = self.ch_names_eeg
+        if self.prep_params["reref_chs"] == "eeg":
+            self.prep_params["reref_chs"] = self.ch_names_eeg
+        self.sfreq = self.raw_eeg.info["sfreq"]
         self.ransac = ransac
         self.random_state = check_random_state(random_state)
 
+    @property
```
sappelhoff, line 118: I have never seen this `property` decorator ... I'll have to look into what it actually does. But it seems fine. :)

sappelhoff (commented):
Sorry I took longer than expected @yjmantilla but I got time for a review.

Looks good to me overall, but I left a few comments, as usual :)

pyprep/prep_pipeline.py
yjmantilla: @christian-oreilly introduced it in the first commit. I have myself seldomly used decorators.  I think that for this particular purpose it adapts well.

sappelhoff: @christian-oreilly could you please explain why you added `@property` here?

sappelhoff:
@yjmantilla when adding code suggestions, you can also first add several suggestions to a "batch" and then commit a batch of suggestions in one go, that's often preferrable to committing each suggestion in one go (at least in my opinion, because I get an email for each commit :slightly_smiling_face: )

for that to work, go to the "files changed" tab in the GitHub PR user interface and click on "add to batch" for as many code suggestions as you think makes sense (e.g., group all "style" suggestions in a batch, then commit ... etc.)

that style of working would be much appreciated in the future :+1:

yjmantilla:
oh sorry, Ill try to remember that in the future, although maybe you said the same in another PR  :/

pyprep/prep_pipeline.py
yjmantilla: in his last comment in #33 he mentioned that : 

> I did not do it in the .raw object directly in case it can have side effect. I made it available through a property that add on the fly the missing channels to the .raw object and return a copy.

If we implement it as a normal attribute I guess said attribute should be set both in the init and at the end of the fit method.

In a way the current implementation does that since it grabs the current raw_eeg and raw_non_eeg .

tests/test_noisy.py
sappelhoff: @yjmantilla do you have time to fix this and fix the CIs? I think then we'd be good to merge this one.

yjmantilla: seems like your LEET guess seed actually works! I think today or during this week the PR will be ready.

pyprep/utils.py
```diff
@@ -86,3 +87,65 @@ def split_list(mylist, chunk_size):
...
+    n_freq_comps : int
+        Number of signal components summed to make a signal.
+    freq_range : list, len==2
+        Signals will contain freqs from this range.
+    scale : float
+        Scaling factor applied to the signal in volts. For example 1e-6 to
+        get microvolts.
+    RNG : np.random.RandomState
+        Random state seed.
+    Returns
```
sappelhoff, line 123:
```suggestion
        Random state seed.
        
    Returns
```

yjmantilla:
sorry for multiple the formatting fixes commits, it is just that I dont have flake8 installed so Im seeing does errors in the remote tests only. Ill see if I cant install it locally now so that there are fewer commits of that sort but I suggest putting that requirement in the development version.

sappelhoff:
> but I suggest putting that requirement in the development version.

it should be there and working:

https://github.com/sappelhoff/pyprep/blob/6bb931c0c8c9739a1dd2b8e047995cb2387b3ede/requirements-dev.txt#L7-L8

but sometimes you need to configure your editor as well so that it properly uses the installed packages and alerts you live while you type

yjmantilla:
hmmm wow, so it has always been there? My bad then, I managed to configure my editor so Im finally seeing those mistakes.

sappelhoff:
argh, now I suggested a line with whitespace to you and that breaks the linter again, in a new, fun way. I'll let you finish without interruptions now :grimacing:

yjmantilla:
That blank line was harder to solve than expected...

I'm not exactly sure what I did to solve it, suddenly by trying different combinations of what apparently looked exactly the same it got solved... It seems like there was some invisible symbol that either was missing or was unneeded there.

sappelhoff (approved):
cool, let's get this in! Thanks a lot everybody!

Once MNE-Python is released (around September 15 https://github.com/mne-tools/mne-python/issues/8150 ) we can add one more feature regarding the windowed mode of the spectrum filter  (https://github.com/sappelhoff/pyprep/issues/35#issuecomment-672702004)

... and then make a new release of pyprep

PR #39, opened 2020-09-03
[API] Remove noisy.py module from pyprep

### LINKED ISSUE #11: noisy.py module should be removed
Before the contributions of @AamnaLawrence @adam2392 and @Nick3151 in #6 there was only the single module `noisy.py` implementing the "find_noisy_channels" functionality.

Now the full PREP is implemented, including a new "find_noisy_channels" functionality, see: https://github.com/sappelhoff/pyprep/blob/master/pyprep/find_noisy_channels.py

`noisy.py` should be removed and the `find_noisy_channels.py` module should be overhauled so that it can be used standalone, without having to run the full PREP.

sappelhoff:
closes #11


yjmantilla:
so sad, noisy.py will be remembered for generations to come.

I do find some of the methods of noisy.py more readable than the current ones. Nevertheless the output of the current implementation should be similar to matlab's prep; I think the John Hopkins team did those tests but Im not sure.

Some of the differences I have spotted are:
- adding  bad_by_SNR , bad_by_dropout attributes and methods.
- find_all_bads function is a bit more obscure since in the current version some of the bad_by_Xmethod are filled because another bad_by_Ymethod calls it internally. I guess that optimizes a bit because no recalculation is done.
- This avoidant of recalculation makes some functions dependent on the state of the class (as in "was the data filtered before by some other class method?" ). In general this mutability of the state make things a bit harder to keep track of. For example the original detrend bug was precisely because of this. 
- by_nan and by_flat where joined into by_nan_flat -> probably because original matlab implementation had it like that
- some parameters where put in the exact number as matlab (ie deviation threshold in 5 rather than 3.29 like noisy.py)
- Some of the methods are a bit more convoluted (ie hf_noise, correlation criteria), I guess because they tried to stay really close as was originally codified in matlab
- ransac is the same apart from the changes I did in the lesser_ram_ransac PR #32

yjmantilla (commented):
All seems ok, although why is the code coverage of the project declining?

EDIT: Seems like the misses are:

- EEG with sample rate beyond 100 in find_bad_by_hfnoise
- if drop[i, j] == 1 in find_bad_by_correlation (not sure what that means though)
- raise raise IOError("Too few channels ...") in find_bad_by_ransac

although that doesn't really account for the decline, maybe it is just rounding errors of some sort.

sappelhoff:
> so sad, noisy.py will be remembered for generations to come.

:laughing: 

> I do find some of the methods of noisy.py more readable than the current ones.

me too (but I am biased :slightly_smiling_face: ), but we can improve the current methods -> first prune the scope of the repo down to something we can maintain -> then add tests -> then refactor / improve

> This avoidant of recalculation makes some functions dependent on the state of the class (as in "was the data filtered before by some other class method?" ). In general this mutability of the state make things a bit harder to keep track of. For example the original detrend bug was precisely because of this.

Yes, I noticed that as well --> this is not very robust and will lead to bugs 

> maybe it is because if one deletes a large chunk of code that was covered then the ratio of lines-covered/total-lines changes and the proportion of the updated coverage is lower.

yes, that's it!

Issue #18, opened 2020-04-06
MNE Multitaper notch filter consumes a lot of RAM.


yjmantilla:
Similar to #17  the MNE notch filter consumes quite a lot of RAM.

For example in my dataset I have 58 channels @ 1000Hz and 5 minutes. This will need somewhere around 9gb of free ram. For personal computers it is quite a lot. Now this would require either looking a way to optimize the notch of MNE and fixing it overthere or make a custom one for pyprep.

Nevertheless in my case my data has already a notch filter for the line noise and maybe some user will already have their data which such filters so I propose for that case to add the feature of optional line noise filtering. This is easily done in prep_pipeline.py by assessing whether line_freqs is empty and in that case to skip directly to the referencing process.

This would look like the following:

```
    if len(self.prep_params["line_freqs"]) != 0:

            # Step 1: 1Hz high pass filtering
            self.EEG_new = removeTrend(self.EEG_raw, sample_rate=self.sfreq)

            # Step 2: Removing line noise
            linenoise = self.prep_params["line_freqs"]

            self.EEG_clean = mne.filter.notch_filter(
                self.EEG_new,
                Fs=self.sfreq,
                freqs=linenoise,
                method="spectrum_fit",
                mt_bandwidth=2,
                p_value=0.01,
            )

            # Add Trend back
            self.EEG = self.EEG_raw - self.EEG_new + self.EEG_clean
            self.raw._data = self.EEG * 1e-6
```

And then the referencing code.

It is important to note that this can be done since the first detrend is temporal and only affects the notch filter since the trend is added back at the end. If we don't do the notch then the first detrend doesn't do anything so thats why it is inside the if.

sappelhoff:
I don't get it --> pyprep currently consumes a lot of run during a single MNE `mne.filter.notch_filter` filter call? How can that be?

yjmantilla:
Yes exactly, in a single filter call it consumes a lot of ram. 

At least speaking for the cleanline method (the filter in matlab) it has to estimate the phase and amplitude of a deterministic sinusoid for each time window for each channel for each expected line frequency harmonics. I noticed this takes quite a lot of time ie 30 min to 1 hour (not sure of the ram consumption in matlab though).

I dont know if the mne multitaper notch filter works exactly the same as the cleanline method of prep but given that in matlab the filter takes that much time it does not surprise me that in mne it requires a lot of ram (and time too).

sappelhoff:
[...]

PR #40, opened 2020-10-09
Adopt windowed approach to spectrum fit notch filtering as the new default

### LINKED ISSUE #35: memory consuming
MemoryError: Unable to allocate 9.75 GiB for an array with shape (6545, 100001) and data type complex128

It just cannot work on a big dataset, which is just 500mb.

yjmantilla:
# PR Description

- Adds the defaults 10s filter length
- Adds filter_kwargs parameter

closes #35 #18 

I dont know if this is what you have in mind...

I'm also a bit busy again ( new exams, deadlines approaching etc) but I managed advance this a bit.

What do you mean by (in #35) 

> And perhaps test whether that'd a good idea?

Originally clean line uses like 2-4 second windows with 50% overlap.

### DIFF
--- a/docs/whats_new.rst
+++ b/docs/whats_new.rst
@@ -32,7 +32,8 @@ Current
 
 Changelog
 ~~~~~~~~~
-
+- It's now possible to pass kwargs arguments to the notch filter inside prep; see the ``filter_kwargs`` parameter by `Yorguin Mantilla`_ (`#40 <https://github.com/sappelhoff/pyprep/pull/40>`_)
+- The default filter length for the spectrum_fit method will be '10s' to fix memory issues, by `Yorguin Mantilla`_ (`#40 <https://github.com/sappelhoff/pyprep/pull/40>`_)
 - Channel types  are now available from a new ``ch_types_all`` attribute, and non-EEG channel names are now available from a new ``ch_names_non_eeg`` attribute from :class:`pyprep.PrepPipeline`, by `Yorguin Mantilla`_ (`#34 <https://github.com/sappelhoff/pyprep/pull/34>`_)
 - Renaming of ``ch_names`` attribute of ``PrepPipeline`` to ``ch_names_all``, by `Yorguin Mantilla`_ (`#34 <https://github.com/sappelhoff/pyprep/pull/34>`_)
 - It's now possible to pass ``'eeg'`` to ``ref_chs`` and ``reref_chs`` parameters to select only eeg channels for referencing, by `Yorguin Mantilla`_ (`#34 <https://github.com/sappelhoff/pyprep/pull/34>`_)
--- a/docs/whats_new.rst
+++ b/docs/whats_new.rst
@@ -32,7 +32,7 @@ Current
 
 Changelog
 ~~~~~~~~~
-- It's now possible to pass kwargs arguments to the notch filter inside prep; see the ``filter_kwargs`` parameter by `Yorguin Mantilla`_ (`#40 <https://github.com/sappelhoff/pyprep/pull/40>`_)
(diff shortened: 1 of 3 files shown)


docs/whats_new.rst
```diff
@@ -32,7 +32,8 @@ Current
 
 Changelog
 ~~~~~~~~~
-
+- It's now possible to pass kwargs arguments to the notch filter inside prep; see the ``filter_kwargs`` parameter by `Yorguin Mantilla`_ (`#40 <https://github.com/sappelhoff/pyprep/pull/40>`_)
```
sappelhoff, line 35: kwargs = keyword arguments, so kwargs arguments = keyword arguments arguments :)

```suggestion
- It's now possible to pass keyword arguments to the notch filter inside prep; see the ``filter_kwargs`` parameter by `Yorguin Mantilla`_ (`#40 <https://github.com/sappelhoff/pyprep/pull/40>`_)
```

pyprep/prep_pipeline.py
```diff
@@ -40,6 +40,9 @@ class PrepPipeline:
         an int, it will be used as a seed for RandomState.
         If None, the seed will be obtained from the operating system
         (see RandomState for details). Default is None.
+    filter_kwargs : dictionary
+        Keywords arguments for the mne.filter.notch_filter function.
+        Note Fs and freqs are already set up by the prep_params dictionary.
```
sappelhoff, line 45:
```suggestion
    filter_kwargs : dict | None
        Optional keywords arguments to be passed on to mne.filter.notch_filter.
        Do not set the "x", Fs", and "freqs" arguments via the filter_kwargs
        parameter, but use the "raw" and "prep_params" parameters instead.
        Can be None to not pass any optional keyword arguments and use the 
        default settings instead.
```

pyprep/prep_pipeline.py
```diff
@@ -133,14 +145,23 @@ def fit(self):
...
+                    p_value=0.01,
+                    filter_length="10s",
+                )
+            else:
+                self.EEG_clean = mne.filter.notch_filter(
+                    self.EEG_new,
+                    Fs=self.sfreq,
+                    freqs=linenoise,
+                    **self.filter_kwargs,
+                )
```
sappelhoff, line 164: instead of this if-else, something more robust would be:

```Python
# set up our defaults ... this goes on line 123
self.filter_kwargs = {
    "x": self.EEG_new,
    "Fs": self.sfreq,
    "freqs": linenoise,
    "method": "spectrum_fit",
    "mt_bandwidth": 2,
    "p_value": 0.01,
    "filter_length": "10s",
}

# now update our defaults with the kwargs
self.filter_kwargs.update(filter_kwargs)

# finally, call mne.filter.notch_filter without if else
self.EEG_clean = mne.filter.notch_filter(**self.filter_kwargs)

```

what do you think @yjmantilla ?

sappelhoff (commented):
thanks a lot @yjmantilla looks good to me overall, but I have one technical remark on how to do this more robustly!

sappelhoff:
> Originally clean line uses like 2-4 second windows with 50% overlap.

okay, that's what I was interested in, thanks. So now we are deviating from that ... but it's better than before, so I am fine with it.

pyprep/prep_pipeline.py
yjmantilla: hmm Im worried about what happens when the user sets for example the method to 'fir' and is expecting the filter length to be the notch filter default 'auto' (so he doesn't set the filter length explicitly) but instead he gests a 10s filter length without him knowing.

sappelhoff: > Also some parameters aren't meaningful when one changes the filter method and supposedly they should be ignored but I guess that could potentially lead to some unknown bugs.

true, good point. We can leave it like you proposed then.

sappelhoff:
Thanks @yjmantilla

PR #44, opened 2020-12-02
Optimize chunk sizes

a-hurst:
# PR Description

On my computer running PyPREP typically slows everything down to a crawl, even causing occasional brief screen blackouts as macOS tries to cope with all its RAM being consumed. Obviously this expected to some extent, but I noticed that it will take as much RAM as possible even when not necessary: e.g. the chunk size being set to 57 for a 64-channel dataset when 32 would have produced the same number of chunks with less RAM consumption.

To address this, I've written a bit of code that figures out the smallest possible chunk size for each possible number of chunks for a file. e.g. for 32 channels, we get:

```python
chunk_sizes = [1, 2, 3, 4, 5, 6, 7, 8, 11, 16, 32]
chunk_counts = [32, 16, 11, 8, 7, 6, 5, 4, 3, 2, 1]
```
When the largest chunk size is too large, it falls back to the next-largest chunk size in the list, so 32 (1 chunk) -> 16 (2 chunks) -> 11 (3 chunks) -> 8 (4 chunks) etc.

In addition to freeing up some system RAM, this should also reduce the likelihood of an overzealous OS killing off the Python process mid-PREP because it's consumed too large a chunk of RAM (has happened to me twice on an 8GB MBP running Catalina, but never on the 24GB iMac running High Sierra or 16GB Hackintosh running Mojave that I've tested PyPREP extensively with)

# Merge Checklist

To merge your PR we need to first take the following points into account:

[contrib-guideline]: https://github.com/sappelhoff/pyprep/blob/master/.github/CONTRIBUTING.md

### DIFF
--- a/pyprep/find_noisy_channels.py
+++ b/pyprep/find_noisy_channels.py
@@ -554,20 +554,21 @@ class NoisyChannels:
         n_chns = chn_pos.shape[0]
 
         # Before running, make sure we have enough memory
-        max_ram_usage = 0.95
+        max_proportion_of_available_ram_to_use = 0.95
         try:
             available_gb = virtual_memory().available * 1e-9
+            available_gb *= max_proportion_of_available_ram_to_use
             needed_gb = (data[:n_chns, :].nbytes * 1e-9) * n_samples
-            needed_gb = needed_gb / max_ram_usage  # adjust for max use
             assert available_gb > needed_gb
         except AssertionError:
+            ram_diff = needed_gb - available_gb
             raise MemoryError(
                 "For given data of shape {} and the requested"
-                " number of {} samples, {} GB or memory would be"
-                " needed but only {} GB are available. You"
-                " could downsample the data or reduce the number"
+                " number of {} samples, {} GB of additional memory"
+                " would be needed. You could close other programs,"
+                " downsample the data, or reduce the number"
                 " of requested samples."
-                "".format(data.shape, n_samples, needed_gb, available_gb)
+                "".format(data.shape, n_samples, ram_diff)
             )
 
         # Memory seems to be fine ...
--- a/pyprep/find_noisy_channels.py
+++ b/pyprep/find_noisy_channels.py
@@ -454,7 +454,16 @@ class NoisyChannels:
 
         print("Executing RANSAC\nThis may take a while, so be patient...")
 
-        chunk_size = self.n_chans_new
+        # Calculate smallest chunk size for each possible chunk count
+        chunk_sizes = []
+        chunk_count = 0
+        for i in range(1, self.n_chans_new + 1):
+            n_chunks = int(np.ceil(self.n_chans_new / i))
+            if n_chunks != chunk_count:
+                chunk_count = n_chunks
+                chunk_sizes.append(i)
+
+        chunk_size = chunk_sizes.pop()
         mem_error = True
         job = list(range(self.n_chans_new))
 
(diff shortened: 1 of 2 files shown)


yjmantilla:
I think this is a better approach of the splitting than mine.

Originally I had
```
                chunk_size = chunk_size - 1  # // 2
```

When I did the PR I compared times and concluded that doing at most channels as possible is indeed considerably faster than splitting each time the chunk_size by 2. So I said "lets do it the fastest way" , and decided to decrease the chunk size by --1 instead of //2 and just commented that (though that comment wasn't really expressive).

>  but I noticed that it will take as much RAM as possible even when not necessary:

It was because I wanted the fastest one :satisfied: , guess it was too much of a greedy algorithm :sweat_smile: . I mean the necessary thing was to be fast. 

Comparing your method and the naive --1 , yours gives a more stable ram consumption. I guess at some points it could make pyprep slower but since it tries to find the fastest method with chunks of roughly the same size I think this is a nice equilibrium point between being fast and using ram in an stable way.

So I guess we are establishing ( in the context of your example): 
- the faster [57,7]  chunks is worst than the bit slower but ram-stable [32,32] chunks.

So if everyone agrees on that statement I say that we can merge this.

Now, just another remark regarding ram compsuntion. Should this line in run_ransac :

```
available_gb = virtual_memory().available * 1e-9
```

Stay like that or be instead a percentage of the ram available (like 70% for example). I guess that if the user is processing and leaving the computer alone is better for it to be 100% but if it is doing something at the same time it would be better a percentage...

What do you think?

yjmantilla (approved):
Only thing missing is to include it in the changelog. I mean I think @sappelhoff would ask me to do that xD.

a-hurst:
I've updated the whats_new and also added a weak implementation of your suggestion: I've made it so it adjusts the `needed_gb` amount, such that the `available_gb` is treated as being 5% smaller than it really is. I'm adjusting the needed_gb amount here instead of `available_gb` directly because otherwise the MemoryError exception text would be confusing.

I guess we can tune the exact proportion of free memory to use, but I think 95% is a good starting place from a stability vs performance perspective (i.e. we aren't losing much). Does that sound good to you?

pyprep/find_noisy_channels.py
```diff
@@ -545,9 +554,11 @@ def run_ransac(
         n_chns = chn_pos.shape[0]
 
         # Before running, make sure we have enough memory
+        max_ram_usage = 0.95
         try:
             available_gb = virtual_memory().available * 1e-9
             needed_gb = (data[:n_chns, :].nbytes * 1e-9) * n_samples
+            needed_gb = needed_gb / max_ram_usage  # adjust for max use
             assert available_gb > needed_gb
```
sappelhoff, line 562: I agree with the point you and @yjmantilla make here, but this is not very readable.

the point is: from overall available RAM, reduce a portion because we anticipate a user to be needed some of that RAM after analyses are started

so I would expect you to tweak `available_gb`, not `needed_gb` (also, dividing with a float < 1 is confusing IMHO)

I suggest:

```suggestion
        # we anticipate users to also need RAM for other things next to pyprep
        max_to_use_from_available_ram = 0.95
        try:
            available_gb = virtual_memory().available * 1e-9
            available_gb *= max_to_use_from_available_ram
            needed_gb = (data[:n_chns, :].nbytes * 1e-9) * n_samples
            assert available_gb > needed_gb
```

sappelhoff (commented):
LGTM apart from one comment I made!

pyprep/find_noisy_channels.py
yjmantilla: I agree with this. Is better that way for readability. Regarding the MemoryError, @a-hurst , you say it is confusing because in available_gb (if multiplied by the percentage) would not be the "real" available gbs or because of what?

sappelhoff: then I'd say we just adjust the error message

a-hurst: I've done this in the latest commit: instead of reporting needed vs available RAM in the exception message, it now just reports the additional required RAM which should avoid confusion. Let me know what you think!

sappelhoff: yes, that's great - thanks!

sappelhoff (approved):

sappelhoff:
Thanks for your first PR to pyprep @a-hurst! may there be many more :tada:

PR #53, opened 2021-02-14
Ransac Comparison Example

### LINKED ISSUE #36: Compare RANSAC in pyprep and autoreject
both the pyprep and the autoreject package have an implementation of RANSAC.

We should make a comparison to:

1. see whether they provide the same results
1. potentially find bugs in one of the implementation
1. check which implementation is more computationally efficient

After the comparison we could decide to drop one implementation and only use the "better" one.

Or we keep using both implementations, but make an example that explicitly compares them, pointing out pros and cons in terms of features.

In any case, this would be an important enterprise for point 1 and 2 in my list alone.

yjmantilla:
# PR Description

Basically a PR for the ransac comparison between autoreject's and pyprep's implementation. 
This is a first version so is a WIP. All ideas welcome for the comparison.

Currently for the real EEG comparison they differ in two channels:

AR: ['C3', 'F7', 'F8', 'FT8', 'T9', 'P7']
PYPREP: ['C3', 'C1', 'F8', 'FT8', 'P7', 'P4']

The reason why is not entirely clear to me. We should maybe investigate further how autoreject's ransac works and whether the input parameters I give him here really do make sense for the comparison. 

I did notice autoreject use an specific way to generate the channels subsets (adds a bias to the seed --> https://github.com/autoreject/autoreject/blob/dd0942438440070876e8ea4796b83ffdd4981600/autoreject/ransac.py#L198 )

We have the correlation matrices of both methods available here for comparison but I have not really used them yet.

Also Autoreject's Ransac is faster at least for these examples. Im not sure yet how is the memory consumption in autoreject. 

The doc's building is failing because autoreject is not installed in whatever executes the example. I don't know where that is configured.

In the final version of this PR it should close #36 

# Merge Checklist


examples/ransac_compare.py
```diff
@@ -0,0 +1,213 @@
...
+
+.. currentmodule:: pyprep
+"""  # noqa: D205 D400
+
+# Authors: Yorguin Mantilla <yjmantilla@gmail.com>
+#
+# License: MIT
+# Based On: run_ransac.py and run_full_prep.py
+
+###############################################################################
```
sappelhoff, line 17:
```suggestion
# %%
```

examples/ransac_compare.py
```diff
@@ -0,0 +1,213 @@
...
+import mne
+from scipy import signal as signal
+from time import perf_counter
+from autoreject import Ransac  # noqa
+from autoreject.utils import interpolate_bads  # noqa
+from mne.utils import check_random_state
+import pyprep.ransac as ransac_pyprep
+
+
+###############################################################################
```
sappelhoff, line 29:
```suggestion
# %%
```

examples/ransac_compare.py
```diff
@@ -0,0 +1,213 @@
...
+]
+# Scale the signal amplitude and add noise.
+X = 2e-5 * np.array(X) + 1e-5 * np.random.random((n_chans, time.shape[0]))
+
+raw = mne.io.RawArray(X, info)
+
+raw.set_montage(montage, verbose=False)
+
+
+###############################################################################
```
sappelhoff, line 75:
```suggestion
# %%
```

examples/ransac_compare.py
```diff
@@ -0,0 +1,213 @@
...
+# Configure RANSAC parameters
+n_samples = 50
+fraction_good = 0.25
+corr_thresh = 0.75
+fraction_bad = 0.4
+corr_window_secs = 5.0
+
+###############################################################################
```
sappelhoff, line 83:
```suggestion
# %%
```

examples/ransac_compare.py
```diff
@@ -0,0 +1,213 @@
...
+print("--- %s seconds ---" % (perf_counter() - start_time))
+
+corr_ar = ransac_ar.corr_
+bad_by_ransac_ar = ransac_ar.bad_chs_
+
+# Check channels that go bad together by correlation (RANSAC)
+print("autoreject bad chs:", bad_by_ransac_ar)
+assert set(bad_ch_names) == set(bad_by_ransac_ar)
+
+###############################################################################
```
sappelhoff, line 114:
```suggestion
# %%
```

examples/ransac_compare.py
```diff
@@ -0,0 +1,213 @@
...
+    channel_wise=False,
+    random_state=random_state,
+)
+print("--- %s seconds ---" % (perf_counter() - start_time))
+
+# Check channels that go bad together by correlation (RANSAC)
+print("pyprep bad chs:", bad_by_ransac_pyprep)
+assert set(bad_ch_names) == set(bad_by_ransac_pyprep)
+
+###############################################################################
```
sappelhoff, line 139:
```suggestion
# %%
```

examples/ransac_compare.py
```diff
@@ -0,0 +1,213 @@
...
+# Now we test the algorithms in a real EEG
+# Let's download some data for testing. Picking the 1st run of subject 4 here.
+data_paths = mne.datasets.eegbci.load_data(subject=4, runs=1, update_path=True)
+fname_test_file = data_paths[0]
+
+###############################################################################
```
sappelhoff, line 145:
```suggestion
# %%
```

examples/ransac_compare.py
```diff
@@ -0,0 +1,213 @@
...
+# The eegbci data has non-standard channel names. We need to rename them:
+mne.datasets.eegbci.standardize(raw)
+
+# Add a montage to the data
+montage_kind = "standard_1005"
+montage = mne.channels.make_standard_montage(montage_kind)
+raw.set_montage(montage)
+
+
+###############################################################################
```
sappelhoff, line 160:
```suggestion
# %%
```

examples/ransac_compare.py
```diff
@@ -0,0 +1,213 @@
...
+print("--- %s seconds ---" % (perf_counter() - start_time))
+
+corr_ar = ransac_ar.corr_
+bad_by_ransac_ar = ransac_ar.bad_chs_
+
+# Check channels that go bad together by correlation (RANSAC)
+print("autoreject bad chs:", bad_by_ransac_ar)
+
+
+###############################################################################
```
sappelhoff, line 191:
```suggestion
# %%
```

examples/ransac_compare.py
```diff
@@ -0,0 +1,213 @@
...
+from scipy import signal as signal
+from time import perf_counter
+from autoreject import Ransac  # noqa
+from autoreject.utils import interpolate_bads  # noqa
+from mne.utils import check_random_state
+import pyprep.ransac as ransac_pyprep
+
+
+###############################################################################
+# Now let's make some arbitrary MNE raw object for demonstration purposes.
```
sappelhoff, line 30: why not use this function for generating a random mne object? https://github.com/sappelhoff/pyprep/blob/8476ea5c1f80c60e564ce8e4c2fca34464515b40/pyprep/utils.py#L93

sappelhoff (commented):
looks like a good start to me :)

I recently found out that in sphinx gallery you can replace the "bar" of `#` with `# %%` signs. I like that second option better, because then editors like VS Code can pick up on that and let you run the example similar to a notebook in the editor.

see here for more info: https://sphinx-gallery.github.io/stable/syntax.html#embed-rst-in-your-example-python-files

regarding the autoreject import, I added it to `requirements-dev.txt` --> that's what contains all dependencies that are needed for testing, building the docs, and development in general.

The dependencies for pyprep are in `setup.cfg`

sappelhoff:
link to example as built by the CI service: https://pyprep--53.org.readthedocs.build/en/53/auto_examples/ransac_compare.html#sphx-glr-auto-examples-ransac-compare-py

examples/ransac_compare.py
```diff
@@ -0,0 +1,213 @@
...
+    fraction_good=fraction_good,
+    corr_thresh=corr_thresh,
+    fraction_bad=fraction_bad,
+    corr_window_secs=corr_window_secs,
+    channel_wise=False,
+    random_state=random_state,
+)
+print("--- %s seconds ---" % (perf_counter() - start_time))
+
+# Check channels that go bad together by correlation (RANSAC)
```
sappelhoff, line 212: re: comment --> "bad by correlation" was a different step to find bad channels, right? RANSAC is a bit more complex than going bad together by correlation, I thought.

yjmantilla: Yeah you are right, is a different step and is more complex. I guess the sole idea is them "going bad together" and using a correlation metric within a RANSAC framework; that is, is not just a simple correlation.


Will leave the original explanation of the paper here for reference so that I can redact a better explanation:

> The predictability criterion also relies on the channel
correlations of the low frequency portion of EEG signals.
Although the bad-by-correlation measure effectively detects most
bad channels, there are some situations in which channels “go
bad together.” We use the RANSAC (random sample consensus)
method (Fischler and Bolles, 1981) to select a random subset of
(so far) good channels to predict the behavior of each (excluded
from subset) channel in small non-overlapping time windows
(5s by default). The implementation is based on functions in
BCILAB (Kothe and Makeig, 2013).

> The bad-by-RANSAC channels behave in a manner poorly
predicted by the other channels. Before applying RANSAC,
we remove channels designated as noisy by other methods.
We then select a random subset of predictor channels for
each channel (25% by default). If not enough channels remain
to form the required subsets, the algorithm terminates. The
RANSAC algorithm uses a method of spherical splines for
estimating scalp potential based on algorithms proposed by
Perrin et al. (1989). Bad-by-RANSAC channels have a correlation
less than a threshold (0.75 by default) with their RANSAC-
predicted time courses on more than a certain fraction of
the windows (0.4 by default). The RANSAC default window
size is 4s.

examples/ransac_compare.py
yjmantilla: Im not yet sure it will work without tweaking the function inside a bit. Since the frequency components are selected randomly from a range inside the function I cannot guarantee easily that the good channels will correlate highly each with each other.

We could make a function to get an arbitrary harmonic signal by summing arbitrary frequencies, amplitudes and phases , though , is it worth it? Something of this sort:

```python
def get_harmonic_signal(freqs,weights,phases,s_rate=1000,start=0,stop=10,withtime=False,time=None):
    """
    Function that returns a harmonic signal given its components.
    Parameters:
        freqs: list of floats
                the frequencies [Hz] of the components
        weights: list of floats
                the weights  of the components
        phases: list of floats
                the phases of the components
        s_rate: float, default=1000
                the sampling rate
        start: float, default=0
                start time
        stop: float, default=10
                stop time
        withtime: bool, default=False
                Whether to return or not
                the time vector of the signal.
        time : np.ndarray, default=None
                1d time vector. If None
                it is created from the start,
                stop and s_rate parameters.
    Returns:
        if withtime:
            time,harmonic_signal
        else:
            harmonic_signal
        harmonic_signal: numpy.ndarray
                the harmonic signal generated
        
    Example:
    >>>s_rate = 1000
    >>>length = s_rate
    >>>freqs_sin = np.array([10,10])
    >>>weights_sin = np.array([1,1])
    >>>phases_sin = np.array([0,90])
    >>>start = 0
    >>>stop = 10
    >>>sin = get_harmonic_signal(freqs_sin,weights_sin,phases_sin,s_rate,start,stop)
    """
    freqs = as_list(freqs)
    weights = as_list(weights)
    phases = as_list(phases)
    if time is None:
        time = np.arange(start, stop, 1/s_rate)
    time = np.squeeze(time)
    harmonic_signal = np.zeros(time.shape[0])
    rad_phases = [phase * np.pi/180 for phase in phases]
    rad_freqs = [2 * np.pi * freq for freq in freqs]
    sinusoid = np.zeros((len(freqs),time.shape[0]))
    for i in range(0,len(freqs)):
        sinusoid[i,:] = weights[i] * np.sin(rad_freqs[i] * time + rad_phases[i])

    harmonic_signal = np.sum(sinusoid,axis=0)
    if withtime == True:
        return time,harmonic_signal
    else:
        return harmonic_signal

```

yjmantilla:
So been delving a bit into the workings of both ransac's.

- pyprep calculates 1 epoch less than AR, i fixed that, was just a limit in a np.arangue
- AR selects particular subset of channels for each (epoch,sample,channel), pyprep just does for (sample,channel)
- AR will iterate through epochs finding the correlations for all channels in each iterations
- Pyprep will iterate through channels finding the correlations all epochs

If we wanted to optimize memory consumption it would be something of the sort of iterating through epochs and channels.

As of now depending on the data pyprep and AR will either consume more memory (I think...,my speculation):
- If there are more channels than epochs --> AR consumes more
- If there are more epochs than channels --> Pyprep consumes more

Pyprep results wont exactly match AR until we manage to work also epoch wise for the prediction, given that the subsets of channels change epoch-wise in Autoreject.

examples/ransac_compare.py
sappelhoff: mmh I take back what I said: It's probably not worth it to make a new big function for that.

sappelhoff:
great work! Thanks, that makes a lot of sense!

> pyprep calculates 1 epoch less than AR, i fixed that, was just a limit in a np.arangue

did you push that commit yet? I can't see it in the changes

> As of now depending on the data pyprep and AR will either consume more memory (I think...,my speculation):

your speculation sounds reasonable :+1: 

> If we wanted to optimize memory consumption it would be something of the sort of iterating through both epochs and channels. Or intelligently select the iteration through the greatest between n_epochs and n_channels

That might be a bit too much optimization, wdyt? the autoreject way already uses quite little memory compared to pyprep's current one.

> In that sense it may be easier to refactor AR to include the channel iteration because they already made the matlab validation.

mh yes, it would be totally fine to have one working RANSAC in python, we don't need different implementations in different packages. But the autoreject RANSAC can only be fit on epochs, it would be nice to be able to fit it on continuous data (even if then under the hood, epochs are being calculated).

yjmantilla:
> did you push that commit yet? I can't see it in the changes

Just did, had to clean up a bit the files since I had experiments in the branch.  So take a look at it, I think it makes sense, it was missing the last possible epoch.

> That might be a bit too much optimization, wdyt? the autoreject way already uses quite little memory compared to pyprep's current one.

I still think it is worthwhile to do. I mean being either of:
- (n_points_1epoch,n_channels,n_samples)
- (n_points_all_epochs,1 channel,n_samples)

Both will consume a lot of ram and for some users it will matter. Guess one could say is not an urgent feature but it is important nonetheless.

> mh yes, it would be totally fine to have one working RANSAC in python, we don't need different implementations in different packages. But the autoreject RANSAC can only be fit on epochs, it would be nice to be able to fit it on continuous data (even if then under the hood, epochs are being calculated).

Yes that fit under the hood was solved by https://github.com/sappelhoff/pyprep/issues/36#issuecomment-737662816 , mainly:

```python
        ############AUTO REJECT####################
        if mode == 'autoreject':
            ransac = Ransac(picks=good_idx,n_resample=n_samples, min_channels=fraction_good, min_corr=corr_thresh, unbroken_time=fraction_bad, n_jobs=njobs, random_state=435656, verbose= 'tqdm')
            info = mne.create_info(self.ch_names_new.tolist(), self.sample_rate, ch_types='eeg', verbose=None)
            # Option 1: Single Epoch
            #EEGData_epoch = self.EEGData[np.newaxis, :, :]
            #epochs = mne.EpochsArray(EEGData_epoch,info)
            
            # Option 2: Epochs of corr_window_secs seconds as windows
            raw = mne.io.RawArray(self.EEGData,info)
            epochs = mne.make_fixed_length_epochs(raw, duration=corr_window_secs, preload=True,reject_by_annotation=False, verbose=None)

            # Either way...
            epochs.set_montage(self.montage)
            epochs_clean = ransac.fit(epochs)
            #print(epochs_clean.bad_chs_)
            self.bad_by_ransac = [i for i in epochs_clean.bad_chs_]
            return None
        ############AUTO REJECT####################
```

Im not how Autoreject does the memory checking but I dont think they do one. So for example may be nice to introduce a verify free ram for them too.

Anyway, what approach do we take? Kill pyprep ransac (so long cowboy...) or keep both and just give the option to run either by a parameter?

I restate here the advantage of AR being already matlab-validated, that is, we could even offer a parameter for it to wrongly calculate the median if the user is a matlab believer.

sappelhoff:
> Both will consume a lot of ram and for some users it will matter. Guess one could say is not an urgent feature but it is important nonetheless.

I think I'd be convinced when I see the first reasonable memory issue for AR ransac :) ... so far I haven't seen any.

> Anyway, what approach do we take? Kill pyprep ransac (so long cowboy...) or keep both and just give the option to run either by a parameter? Do direct pull request to AR if we need changes?

it would probably be more sustainable to kill pyprep ransac, and stay with the wrapped AR ransac within pyprep. The wrapping then allows us to build some functionality on top of AR ransac that would be out of scope in the AR package (such as support for continuous, as opposed to epochs). -- of course, core "enhancements" could then be done as PRs to the AR package.


> I restate here the advantage of AR being already matlab-validated, that is, we could even offer a parameter for it to wrongly calculate the median if the user is a matlab believer.

yes, but is there an open comparison of AR-ransac and PREP-ransac? Or is this all buried in Mainak's PhD work? :-)

yjmantilla:
> I think I'd be convinced when I see the first reasonable memory issue for AR ransac :) ... so far I haven't seen any.
Lets see how it goes 😬 

> yes, but is there an open comparison of AR-ransac and PREP-ransac? Or is this all buried in Mainak's PhD work? :-)

Yeah, i guess thats buried somewhere, although Mainak has left hints along the comments he has made in pyprep's issues.

So what i take from this is that this PR will mutate to:

1 - Implement AR ransac inside pyprep
2 - Example will now be between MATLAB RANSAC and AR RANSAC rather than pyprep's RANSAC.

Is this correct?

sappelhoff:
>  Implement AR ransac inside pyprep

yes (=adding autoreject as a dependency), and let's keep pyprep ransac around for some more time --- once we are sure that the AR ransac works fine for our needs, we can remove pyprep ransac.

> Example will now be between MATLAB RANSAC and AR RANSAC rather than pyprep's RANSAC.

yes, either that, or comparing all three --> but in the long term then rather only AR ransac :+1:

sappelhoff:
tests are failing here because I wanted to pass a `np.random.RandomState` object to autoreject's RANSAC. Opened an issue about that here: https://github.com/autoreject/autoreject/issues/218

On a different note: This PR is now very old and A LOT of stuff has happened in the meantime, so my opinion from https://github.com/sappelhoff/pyprep/pull/53#issuecomment-786504718 is no longer up to date.

Back then I thought it'd be good to start depending on autoreject's RANSAC, modify it to our needs, and part by part remove pyprep's RANSAC.

Now I think we should keep our ransac with all of @a-hurst's fixes. 

Still I think it would be nice to have this example of comparing pyprep's and autoreject's RANSAC. So I'd say, let's review / improve this a bit and then merge it. What do you think @a-hurst @yjmantilla ?

One last important note: I merged `master` into this branch in https://github.com/sappelhoff/pyprep/pull/53/commits/debef8ca6b3afdb76b5f8bd1a8bd260c4f134f8d and that "removed" a fix that @yjmantilla added and described in https://github.com/sappelhoff/pyprep/pull/53#issuecomment-784741114

should we add that back? Could you have a look how that fits in with your recent changes @a-hurst ?

a-hurst:
Wow, looking at the commit that syncs this up with the current version you can really see the extent of the work that's been done on pyprep in the past few months. Nice!

> So I'd say, let's review / improve this a bit and then merge it. What do you think @a-hurst @yjmantilla?

Yep, this definitely seems useful! I still need to read this PR over properly so it may have this already, but it would be great to have a section in the example that compares/contrasts the usage of PyPREP vs AutoReject (i.e., types of noise detection, full file vs per-epoch) so that understanding when/where to use each is clearer to newcomers (I remember getting confused by this myself last year when first setting up my EEG pipeline).

I can contribute a bit to this since I've stared into the depths of PyPREP long enough to roughly know how it works/behaves and I don't mind writing docs, I'd just need to get a handle on how AutoReject works enough to compare/contrast. I also wonder about whether AutoReject and PyPREP might be usable together once PyPREP's final interpolation of bad channels is made optional, i.e. using PyPREP for adaptive line noise removal and robust re-referencing, but using AutoReject for final bad channel detection/interpolation post-reference to avoid unnecessary interpolation of channels that are only bad during a certain period of time.

>One last important note: I merged master into this branch in debef8c and that "removed" a fix that @yjmantilla added and described in #53 (comment)
>
>should we add that back? Could you have a look how that fits in with your recent changes @a-hurst ?

Are you saying that this PR also had an implementation of window-wise chunking for RANSAC, that got reverted with my current implementation? I looked through the commit you linked but couldn't find it: ransac.py looks functionally identical to how it did before I started my RANSAC rewriting/refactoring. Can you point me to the lines in question?

sappelhoff:
I meant specifically this commit: https://github.com/sappelhoff/pyprep/pull/53/commits/57043ac89e1e6a0dccdfe412a1178acc60e21dda

note the `correlation_frames + 1`

which we don't have in `master`

https://github.com/sappelhoff/pyprep/blob/4614ca01f4e84a93c04fdc729185306d715a1a7c/pyprep/ransac.py#L168-L175

sappelhoff:
> I can contribute a bit to this since I've stared into the depths of PyPREP long enough to roughly know how it works/behaves and I don't mind writing docs, I'd just need to get a handle on how AutoReject works enough to compare/contrast.

pretty sure that @yjmantilla would welcome your help on this - as would I :-)

>  I also wonder about whether AutoReject and PyPREP might be usable together once PyPREP's final interpolation of bad channels is made optional, i.e. using PyPREP for adaptive line noise removal and robust re-referencing, but using AutoReject for final bad channel detection/interpolation post-reference to avoid unnecessary interpolation of channels that are only bad during a certain period of time.

could be a nice future example / documentation (separate from this thread / PR)

examples/ransac_comparison_autoreject.py
```diff
@@ -0,0 +1,212 @@
...
+    corr_thresh=corr_thresh,
+    frac_bad=fraction_bad,
+    corr_window_secs=corr_window_secs,
+    channel_wise=False,
+    random_state=rng,
+)
+print("--- %s seconds ---" % (perf_counter() - start_time))
+
+# Check channels that go bad together by RANSAC
+print("pyprep bad chs:", bad_by_ransac_pyprep)
```
sappelhoff, line 212: we should add one more code block below to easily show:

- chs bad that are equal between pyprep and AR
- chs bad only AR
- chs bad only pyprep

a-hurst:
> I meant specifically this commit: [57043ac](https://github.com/sappelhoff/pyprep/commit/57043ac89e1e6a0dccdfe412a1178acc60e21dda)
> 
> note the `correlation_frames + 1`
> 
> which we don't have in `master`

Ahh, I see. Presumably that matches up PyPREP's RANSAC with AutoReject's, but that change would diverge from MATLAB PREP so we'd need to wrap it in `matlab_strict` if we wanted to add that change.

My understanding was that PREP/PyPREP don't use that last correlation window because it's at the end of the file and would thus be shorter than the rest of the windows. In AR, making use of that last window makes sense since you want all the data per epoch that you can get, but for PyPREP I think it makes sense to ignore it (e.g. the last window might only be 100 ms long, but would be weighted the same as a full 5 second window in determining whether a channel is bad or not). Does that make sense?

sappelhoff:
ah yes, that makes sense. A feature for -non-matlab-strict pyprep may be to include the final window if it is only X% shorter than the others, where X could be set to something like 20% ....  and if it's more than 20% shorter, don't calculate anything based on it.

it would be good to add that explanation within the codebase in some way, so that future devs can easily know the reasoning.

yjmantilla:
@sappelhoff @a-hurst 

Sorry for the late reply. You guys have been really working hard on pyprep and currently I cant keep up with your fast work hehe.

>Now I think we should keep our ransac with all of @a-hurst's fixes.

I agree, after all that was the initial idea.

>Still I think it would be nice to have this example of comparing pyprep's and autoreject's RANSAC. So I'd say, let's review / improve this a bit and then merge it. What do you think @a-hurst @yjmantilla ?

I agree, although it wouldnt be high priority imo. 

>One last important note: I merged master into this branch in debef8c and that "removed" a fix that @yjmantilla added and described in #53 (comment)
>Ahh, I see. Presumably that matches up PyPREP's RANSAC with AutoReject's, but that change would diverge from MATLAB PREP so we'd need to wrap it in matlab_strict if we wanted to add that change.

Yes,as @a-hurst  said, this was a "matching" I did so that the matrix of correlations had the same shape (and presumably the windows in both implementations of pyprep and ar ransac were the same). That way I could numerically compare the matrix of correlations (or whatever matrix got outside of ransac, I dont quite remember).

>My understanding was that PREP/PyPREP don't use that last correlation window because it's at the end of the file and would thus be shorter than the rest of the windows. In AR, making use of that last window makes sense since you want all the data per epoch that you can get, but for PyPREP I think it makes sense to ignore it (e.g. the last window might only be 100 ms long, but would be weighted the same as a full 5 second window in determining whether a channel is bad or not). Does that make sense?

Nice logic. Maybe for the comparison of the correlation matrix between Pyprep's and AR's RANSAC one can just use every window but the last on the AR's Ransac side. I wouldnt know at this moment if the windows align exactly u.u , but I think at the time of this PR they did. After all is a matter of checking the window chunking logic and maybe numerically checking they align if they both have that info in an array like for example the 'correlation offsets' one.

>it would be good to add that explanation within the codebase in some way, so that future devs can easily know the reasoning.

I agree.

sappelhoff:
closing this now as i brush up the repo a little :)

thanks for the efforts and discussion back then and i hope you are fine! 👋

PR #57, opened 2021-04-08
Add MATLAB-style implementations of the quantile and IQR functions

### LINKED ISSUE #56: Bad channels by correlation differ between PyPREP and MATLAB PREP
In my recent comparison efforts, I noticed that the numbers of initial bad channels by correlation sometimes differed slightly between PyPREP and MATLAB PREP. Since this code doesn't involve any random number generation, it should always be consistent between runs and also between implementations.

Anyway, I did some digging with breakpoints and think I've figured it out. It has to do with [these lines here](https://github.com/sappelhoff/pyprep/blob/8476ea5c1f80c60e564ce8e4c2fca34464515b40/pyprep/find_noisy_channels.py#L320-L332):

```python
        for k in range(0, w_correlation):
            eeg_portion = np.transpose(np.squeeze(EEG_new_win[:, :, k]))
            data_portion = np.transpose(np.squeeze(data_win[:, :, k]))
            window_correlation = np.corrcoef(np.transpose(eeg_portion))
            abs_corr = np.abs(
                np.subtract(window_correlation, np.diag(np.diag(window_correlation)))
            )
            channel_correlation[k, :] = np.quantile(abs_corr, 0.98, axis=0)
            noiselevels[k, :] = np.divide(
                robust.mad(np.subtract(data_portion, eeg_portion), c=1),
                robust.mad(eeg_portion, c=1),
            )
            channel_deviations[k, :] = 0.7413 * iqr(data_portion, axis=0)
```
Using the same input data, we see that `eeg_portion`, `data_portion`, `window_correlation`, `abs_corr`, and `noiselevels` are all identical between PyPREP and MatPREP, but `channel_correlation` and `channel_deviations` both differ.

The reason for this is that apparently [MATLAB's quantile functions work slightly differently than Numpy's and SciPy's](https://stackoverflow.com/a/58424860): instead of having quantiles be restricted to the range [0, 1], they get restricted to the range [0.5 / n, (n - 0.5) / n], with n being the number of samples passed to the quantile function. This is apparently equivalent to treating the provided values as a sample of an unknown population rather than a full population themselves. After hacking together some new `quantile` and `iqr` functions that account for this, the `channel_correlation` and `channel_deviations` data both match the MATLAB values!

Anyway, I'll put together a PR to address this, but I figure there should be a discussion of which quantile logic actually makes the most sense for this use-case. If Numpy's default logic is actually more reasonable/justifiable here, we can make the new quantile functions optional with a `match_matlab` flag  or parameter of some kind indicating strict adherence to the original PREP logic. If MATLAB's logic makes more sense, I'm assuming we should just fully replace `np.quantile` and `iqr` with their modified equivalents?

a-hurst:
# PR Description

Closes #56. As mentioned in that issue, MATLAB's quantile and IQR functions calculate quantiles slightly differently than Numpy and SciPy, leading to higher sensitivity to differences in channel correlation than in the original MATLAB PREP. This PR adds new `mat_quantile` and `mat_iqr` functions that match MATLAB's quantile logic and make PyPREP's internal correlation matrix for `NoisyChannels.find_bad_by_correlation` match the values in MATLAB PREP's equivalent matrix exactly.

### TODO

- I wasn't sure whether we'd want these functions to be part of the public API or private (i.e. `_mat_irq` and `_mat_quantile`) so I held off on writing docstrings.
- I also wasn't sure what we wanted to do here for unit testing. For the sake of this PR, would individual tests for these two functions be fine, or should we add full-pipeline testing of these changes as well? Eventually we'll want to modify the CI tests to do proper comparisons between MATLAB and PyPREP output values for each stage of NoisyChannels, but that would require figuring out some good test data we could use for both pipelines that wouldn't demand too much in terms of RAM or CPU, so I think that's a bigger problem to tackle in a separate PR.
- I need to update the changelog (which sort of depends on whether we want the functions to be public or private).

# Merge Checklist

### DIFF
--- a/docs/whats_new.rst
+++ b/docs/whats_new.rst
@@ -42,6 +42,7 @@ Bug
 ~~~
 - Fixed RANSAC to give consistent results with a fixed seed across different chunk sizes, by `Austin Hurst`_ and `Yorguin Mantilla`_ (:gh:`43`)
 - Fixed "bad channel by flat" threshold in :meth:`NoisyChannels.find_bad_by_nan_flat` to be consistent with MATLAB PREP, by `Austin Hurst`_ (:gh:`60`)
+- Changed "bad channel by deviation" and "bad channel by correlation" detection code in :class:`NoisyChannels` to produce identical results to MATLAB PREP, by `Austin Hurst`_ (:gh:`57`)
 
 API
 ~~~
--- a/docs/whats_new.rst
+++ b/docs/whats_new.rst
@@ -42,7 +42,7 @@ Bug
 ~~~
 - Fixed RANSAC to give consistent results with a fixed seed across different chunk sizes, by `Austin Hurst`_ and `Yorguin Mantilla`_ (:gh:`43`)
 - Fixed "bad channel by flat" threshold in :meth:`NoisyChannels.find_bad_by_nan_flat` to be consistent with MATLAB PREP, by `Austin Hurst`_ (:gh:`60`)
-- Changed "bad channel by deviation" and "bad channel by correlation" detection code in :class:`NoisyChannels` to produce identical results to MATLAB PREP, by `Austin Hurst`_ (:gh:`57`)
+- Changed "bad channel by deviation" and "bad channel by correlation" detection code in :class:`NoisyChannels` to compute IQR and quantiles in the same manner as MATLAB, thus producing identical results to MATLAB PREP, by `Austin Hurst`_ (:gh:`57`)
 
 API
 ~~~
--- a/pyprep/utils.py
+++ b/pyprep/utils.py
@@ -5,6 +5,7 @@ from cmath import sqrt
 import mne
 import numpy as np
 import scipy.interpolate
+from scipy.stats import iqr
 from psutil import virtual_memory
 
 
@@ -18,6 +19,28 @@ def _set_diff(list1, list2):
 
 def _intersect(list1, list2):
     return list(set(list1).intersection(set(list2)))
+
+
+def mat_quantile(arr, q, axis=None):
+    # MATLAB calculates quantiles using different logic than Numpy: whereas
+    # Numpy treats the provided values as a whole population, MATLAB treats
+    # the input array as a sample from a population of unknown size.
+    # This function adjusts the quantiles to mimic MATLAB's behaviour.
+    q = np.asarray(q, dtype=np.float64)
+    n = len(arr)
+    q_adj = ((q - 0.5) * n / (n - 1)) + 0.5
+    return np.quantile(arr, np.clip(q_adj, 0, 1), axis=axis)
+
+
+def mat_iqr(arr, axis=None):
+    # MATLAB calculates IQRs using different logic than Numpy: whereas
+    # Numpy treats the provided values as a whole population, MATLAB treats
+    # the input array as a sample from a population of unknown size.
+    # This function adjusts the IQR to mimic MATLAB's behaviour.
+    iqr_q = np.asarray([25, 75], dtype=np.float64)
+    n = len(arr)
+    iqr_adj = ((iqr_q - 50) * n / (n - 1)) + 50
+    return iqr(arr, rng=np.clip(iqr_adj, 0, 100), axis=axis)
 
 
 def filter_design(N_order, amp, freq):
--- a/pyprep/utils.py
+++ b/pyprep/utils.py
@@ -22,7 +22,7 @@ def _intersect(list1, list2):
 
 
 def _mat_quantile(arr, q, axis=None):
-    """Calculate the numeric value at quantile (q) for a given distribution.
+    """Calculate the numeric value at quantile (`q`) for a given distribution.
 
     Parameters
     ----------
--- a/pyprep/utils.py
+++ b/pyprep/utils.py
@@ -22,8 +22,7 @@ def _intersect(list1, list2):
 
 
 def _mat_quantile(arr, q, axis=None):
-    """Calculate the numeric value at quantile (q) for a given distribution of
-    values, using the same formula as MATLAB and Octave.
+    """Calculate the numeric value at quantile (q) for a given distribution.
 
     Parameters
     ----------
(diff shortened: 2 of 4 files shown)


yjmantilla:
+1 for individual tests , at least at this stage. It would be nice to see how it affects the final result. If you do a test comparing the saved matlab result (of only these functions) against this implementation and they come out as working better than before then I think that is enough; we are getting closer to matlab.

+1 to private functions, I think this is too deep inside the code to be public, nor we are developing specifically these statistical functions.

sappelhoff:
> +1 to private functions, I think this is too deep inside the code to be public, nor we are developing specifically these statistical functions.

agreed

> I wasn't sure whether we'd want these functions to be part of the public API or private (i.e. _mat_irq and _mat_quantile) so I held off on writing docstrings.

docstrings are almost *always* a good idea - even if it's not public-facing, the documentation will help us (the developers) in the future. Following the numpydoc convention makes the whole thing well structured and easy to read for us --> even if it does not end up being rendered as html on our docs page.


> As you say a full pipeline test matlab vs python would be another PR . Thats a separate issue on its own.

agreed

pyprep/utils.py
```diff
@@ -20,6 +21,28 @@ def _intersect(list1, list2):
...
+    n = len(arr)
+    q_adj = ((q - 0.5) * n / (n - 1)) + 0.5
+    return np.quantile(arr, np.clip(q_adj, 0, 1), axis=axis)
+
+
+def mat_iqr(arr, axis=None):
+    # MATLAB calculates IQRs using different logic than Numpy: whereas
+    # Numpy treats the provided values as a whole population, MATLAB treats
+    # the input array as a sample from a population of unknown size.
+    # This function adjusts the IQR to mimic MATLAB's behaviour.
```
sappelhoff, line 39: The comments you make here could well go into a `Notes` section of a numpydoc string

```python
"""
...

Notes
-----
MATLAB calculates quantiles using different logic than Numpy: where ....
"""
```

a-hurst:
@yjmantilla @sappelhoff Thanks for reviewing this! I've updated the pull request accordingly, and have also rebased to the current master branch.

tests/test_utils.py
```diff
@@ -0,0 +1,37 @@
...
+       rng(435656);
+       tst = rand(100, 3);
+
+       % Calculate IQR and 0.98 quantile for test data
+       quantile(tst, 0.98);
+       iqr(tst);
+
+    """
+    # Generate test data
+    np.random.seed(435656)
```
sappelhoff, line 24: so here the numpy seed corresponds to the matlab seed? (looking at the matlab codeblock `rng(435656);` in the docstring, line 15)

pyprep/utils.py
```diff
@@ -20,6 +21,71 @@ def _intersect(list1, list2):
...
+
+
+def _mat_iqr(arr, axis=None):
+    """Calculate the inter-quartile range (IQR) for a given distribution.
+
+    Parameters
+    ----------
+    arr : np.ndarray
+        Input array containing samples from the distribution to summarize.
+    axis : {int, tuple of int, None}, optional
```
sappelhoff, line 67: where did you get the idea to document the different options in curly brackets?

I usually do: `axis : int | tuple of int | None`, and then write what it defaults to in the description string below (not writing "optional" explicitly)

which doesn't mean that that's better (hence my question).

pyprep/utils.py
```diff
@@ -20,6 +21,71 @@ def _intersect(list1, list2):
     return list(set(list1).intersection(set(list2)))
 
 
+def _mat_quantile(arr, q, axis=None):
+    """Calculate the numeric value at quantile (q) for a given distribution.
```
sappelhoff, line 25: Just a very nitpicky and unimportant note on numpydoc style: when you describe parameters in the docstring, you should wrap them in single backticks, like:

```suggestion
    """Calculate the numeric value at quantile (`q`) for a given distribution.
```

See the "4. parameters" section here: https://numpydoc.readthedocs.io/en/latest/format.html#sections

docs/whats_new.rst
```diff
@@ -42,6 +42,7 @@ Bug
 ~~~
 - Fixed RANSAC to give consistent results with a fixed seed across different chunk sizes, by `Austin Hurst`_ and `Yorguin Mantilla`_ (:gh:`43`)
 - Fixed "bad channel by flat" threshold in :meth:`NoisyChannels.find_bad_by_nan_flat` to be consistent with MATLAB PREP, by `Austin Hurst`_ (:gh:`60`)
+- Changed "bad channel by deviation" and "bad channel by correlation" detection code in :class:`NoisyChannels` to produce identical results to MATLAB PREP, by `Austin Hurst`_ (:gh:`57`)
```
sappelhoff, line 45: I think you can be tiny bit more explicit/verbose here: *what* was changed: Numpy quantile and IQR functions were adjusted to mirror Matlab's quantile and IQR --> in order to produce identical results to MATLAB PREP.

sappelhoff (approved):
looks great :) some minor questions/clarifications

tests/test_utils.py
a-hurst: Yep, that's correct: 435656 is the default MATLAB PREP seed.

pyprep/utils.py
a-hurst: Oh, I just copy/pasted that bit directly from the [docstring of `numpy.quantile`](https://github.com/numpy/numpy/blob/v1.20.0/numpy/lib/function_base.py#L3827-L3931) since that arguments getting passed unmodified to that function anyway. No rationale whatsoever on my part, just figured I'd go with whatever the Numpy maintainers thought was the correct format (I'm new to Numpy docstrings, I've previously only used Google-style ones for Python).

If you'd like I can change it to you preferred format for the sake of consistency across PyPREP!

pyprep/utils.py
a-hurst: Good catch, will fix.

pyprep/utils.py
sappelhoff: ah thanks, I read the [numpydoc styleguide](https://numpydoc.readthedocs.io/en/latest/format.html) again, and it seems like the curly brackets is advertised. I think I picked up the `|` style from MNE-Python. Let's stick with the curly braces here and in general slowly convert to that style whenever we touch (git diff) a line that still uses the `|` style in the remaining codebase.

tests/test_utils.py
sappelhoff: perfect, thanks!

docs/whats_new.rst
a-hurst: Hopefully addressed in my latest commit.

sappelhoff:
Thanks @a-hurst  :-) on a roll with these PRs

PR #66, opened 2021-04-16
Add window-wise RANSAC to reduce RAM requirements

a-hurst:
# PR Description

(Updated) This PR does a few things:

- Adds support for window-wise RANSAC and makes it the default method
- Reworks channel-wise RANSAC to use the same pre-generated interpolation matrices as window-wise, which speeds things up and reduces the amount of divergent/redundant code between methods
- Reworks/refactors the channel-wise RANSAC functions a fair bit to better match the function names, argument names, variable names, and docstrings of their window-wise equivalence, for easier comparison and maintainability
- Adds a progress-printing function for window-wise RANSAC
- Overhauls the RANSAC unit tests considerably, including the addition of explicit comparison of the window-wise/channel-wise correlation matrices for both regular and MATLAB-strict RANSAC to make sure we get identical results for the same seed between methods (and that MATLAB strict values differ from non-strict ones as expected).

### Currently Missing
- [x] Progress indicators for RANSAC
- [x] Proper API for `matlab_strict` mode (handled in earlier PR)
- [ ] Any sort of optimization/parallelization for additional speed (I have some ideas, but will save for a future PR since this one's already pretty big)
- [x] Any sort of memory checking or channel chunking support (added memory check at start)
- [x] A clear idea of what to do about MATLAB PREP's weird correlation function (handled in earlier PR)

### Other Notes
- For my test file (`sub_002/ses-01` from that open meditation dataset), window-wise RANSAC is about 36% slower than the existing RANSAC approach (192 vs 141 seconds). However, RAM usage during RANSAC is now much lower (maxed out at 2.85 GB), which is nice but also suggests that window-wise RANSAC could be improved to take better advantage of RAM on beefier systems.

# Merge Checklist

### DIFF
--- a/pyprep/find_noisy_channels.py
+++ b/pyprep/find_noisy_channels.py
@@ -479,6 +479,7 @@ class NoisyChannels:
             fraction_bad,
             corr_window_secs,
             channel_wise,
+            False,
             self.random_state,
             self.matlab_strict,
         )
--- a/pyprep/find_noisy_channels.py
+++ b/pyprep/find_noisy_channels.py
@@ -144,22 +144,42 @@ class NoisyChannels:
             )
         return bads
 
-    def find_all_bads(self, ransac=True):
+    def find_all_bads(self, ransac=True, channel_wise=False, max_chunk_size=None):
         """Call all the functions to detect bad channels.
 
         This function calls all the bad-channel detecting functions.
 
         Parameters
         ----------
-        ransac : bool
-            To detect channels by ransac or not.
+        ransac : bool, optional
+            Whether RANSAC should be for bad channel detection, in addition to
+            the other methods. RANSAC can detect bad channels that other methods
+            are unable to catch, but also slows down noisy channel detection
+            considerably. Defaults to ``True``.
+        channel_wise : bool, optional
+            Whether RANSAC should predict signals for whole chunks of channels
+            at once instead of predicting signals for each RANSAC window
+            individually. Channel-wise RANSAC generally has higher RAM demands
+            than window-wise RANSAC (especially if `max_chunk_size` is
+            ``None``), but can be faster on systems with lots of RAM to spare.
+            Has no effect if not using RANSAC. Defaults to ``False``.
+        max_chunk_size : {int, None}, optional
+            The maximum number of channels to predict at once during
+            channel-wise RANSAC. If ``None``, RANSAC will use the largest chunk
+            size that will fit into the available RAM, which may slow down
+            other programs on the host system. If using window-wise RANSAC
+            (the default) or not using RANSAC at all, this parameter has no
+            effect. Defaults to ``None``.
 
         """
         self.find_bad_by_nan_flat()
         self.find_bad_by_deviation()
         self.find_bad_by_SNR()
         if ransac:
-            self.find_bad_by_ransac()
+            self.find_bad_by_ransac(
+                channel_wise=channel_wise,
+                max_chunk_size=max_chunk_size
+            )
 
     def find_bad_by_nan_flat(self):
         """Detect channels that appear flat or have NaN values."""
--- a/pyprep/find_noisy_channels.py
+++ b/pyprep/find_noisy_channels.py
@@ -152,17 +152,20 @@ class NoisyChannels:
         Parameters
         ----------
         ransac : bool, optional
-            Whether RANSAC should be for bad channel detection, in addition to
-            the other methods. RANSAC can detect bad channels that other methods
-            are unable to catch, but also slows down noisy channel detection
-            considerably. Defaults to ``True``.
+            Whether RANSAC should be used for bad channel detection, in addition
+            to the other methods. RANSAC can detect bad channels that other
+            methods are unable to catch, but also slows down noisy channel
+            detection considerably. Defaults to ``True``.
         channel_wise : bool, optional
-            Whether RANSAC should predict signals for whole chunks of channels
-            at once instead of predicting signals for each RANSAC window
-            individually. Channel-wise RANSAC generally has higher RAM demands
-            than window-wise RANSAC (especially if `max_chunk_size` is
-            ``None``), but can be faster on systems with lots of RAM to spare.
-            Has no effect if not using RANSAC. Defaults to ``False``.
+            Whether RANSAC should predict signals for chunks of channels over the
+            entire signal length ("channel-wise RANSAC", see `max_chunk_size`
+            parameter). If ``False``, RANSAC will instead predict signals for all
+            channels at once but over a number of smaller time windows instead of
+            over the entirem signal length ("window-wise RANSAC"). Channel-wise
+            RANSAC generally has higher RAM demands than window-wise RANSAC
+            (especially if `max_chunk_size` is ``None``), but can be faster on
+            systems with lots of RAM to spare. Has no effect if not using RANSAC.
+            Defaults to ``False``.
         max_chunk_size : {int, None}, optional
             The maximum number of channels to predict at once during
             channel-wise RANSAC. If ``None``, RANSAC will use the largest chunk
@@ -468,12 +471,14 @@ class NoisyChannels:
             The duration (in seconds) of each RANSAC correlation window. Defaults
             to 5 seconds.
         channel_wise : bool, optional
-            Whether RANSAC should predict signals for whole chunks of channels
-            at once instead of predicting signals for each RANSAC window
-            individually. Channel-wise RANSAC generally has higher RAM demands
-            than window-wise RANSAC (especially if `max_chunk_size` is
-            ``None``), but can be faster on systems with lots of RAM to spare.
-            Defaults to ``False``.
+            Whether RANSAC should predict signals for chunks of channels over the
+            entire signal length ("channel-wise RANSAC", see `max_chunk_size`
+            parameter). If ``False``, RANSAC will instead predict signals for all
+            channels at once but over a number of smaller time windows instead of
+            over the entirem signal length ("window-wise RANSAC"). Channel-wise
+            RANSAC generally has higher RAM demands than window-wise RANSAC
+            (especially if `max_chunk_size` is ``None``), but can be faster on
+            systems with lots of RAM to spare. Defaults to ``False``.
         max_chunk_size : {int, None}, optional
             The maximum number of channels to predict at once during
             channel-wise RANSAC. If ``None``, RANSAC will use the largest chunk
--- a/pyprep/find_noisy_channels.py
+++ b/pyprep/find_noisy_channels.py
@@ -161,7 +161,7 @@ class NoisyChannels:
             entire signal length ("channel-wise RANSAC", see `max_chunk_size`
             parameter). If ``False``, RANSAC will instead predict signals for all
             channels at once but over a number of smaller time windows instead of
-            over the entirem signal length ("window-wise RANSAC"). Channel-wise
+            over the entire signal length ("window-wise RANSAC"). Channel-wise
             RANSAC generally has higher RAM demands than window-wise RANSAC
             (especially if `max_chunk_size` is ``None``), but can be faster on
             systems with lots of RAM to spare. Has no effect if not using RANSAC.
@@ -475,7 +475,7 @@ class NoisyChannels:
             entire signal length ("channel-wise RANSAC", see `max_chunk_size`
             parameter). If ``False``, RANSAC will instead predict signals for all
             channels at once but over a number of smaller time windows instead of
-            over the entirem signal length ("window-wise RANSAC"). Channel-wise
+            over the entire signal length ("window-wise RANSAC"). Channel-wise
             RANSAC generally has higher RAM demands than window-wise RANSAC
             (especially if `max_chunk_size` is ``None``), but can be faster on
             systems with lots of RAM to spare. Defaults to ``False``.
--- a/pyprep/find_noisy_channels.py
+++ b/pyprep/find_noisy_channels.py
@@ -409,6 +409,7 @@ class NoisyChannels:
         fraction_bad=0.4,
         corr_window_secs=5.0,
         channel_wise=False,
+        max_chunk_size=None,
     ):
         """Detect channels that are not predicted well by other channels.
 
@@ -447,10 +448,18 @@ class NoisyChannels:
             The duration (in seconds) of each RANSAC correlation window. Defaults
             to 5 seconds.
         channel_wise : bool, optional
-            Whether RANSAC should be performed one channel at a time (lower RAM
-            demands) or in chunks of as many channels as can fit into the
-            currently available RAM (faster). Defaults to ``False`` (i.e., using
-            the faster method).
+            Whether RANSAC should predict signals for whole chunks of channels
+            at once instead of predicting signals for each RANSAC window
+            individually. Channel-wise RANSAC generally has higher RAM demands
+            than window-wise RANSAC (especially if `max_chunk_size` is
+            ``None``), but can be faster on systems with lots of RAM to spare.
+            Defaults to ``False``.
+        max_chunk_size : {int, None}, optional
+            The maximum number of channels to predict at once during
+            channel-wise RANSAC. If ``None``, RANSAC will use the largest chunk
+            size that will fit into the available RAM, which may slow down
+            other programs on the host system. If using window-wise RANSAC
+            (the default), this parameter has no effect. Defaults to ``None``.
 
         References
         ----------
@@ -479,7 +488,7 @@ class NoisyChannels:
             fraction_bad,
             corr_window_secs,
             channel_wise,
-            False,
+            max_chunk_size,
             self.random_state,
             self.matlab_strict,
         )
--- a/pyprep/prep_pipeline.py
+++ b/pyprep/prep_pipeline.py
@@ -38,6 +38,19 @@ class PrepPipeline:
     ransac : bool, optional
         Whether or not to use RANSAC for noisy channel detection in addition to
         the other methods in :class:`~pyprep.NoisyChannels`. Defaults to True.
+    channel_wise : bool, optional
+        Whether RANSAC should predict signals for whole chunks of channels at
+        once instead of predicting signals for each RANSAC window
+        individually. Channel-wise RANSAC generally has higher RAM demands than
+        window-wise RANSAC (especially if `max_chunk_size` is ``None``), but can
+        be faster on systems with lots of RAM to spare.nHas no effect if not
+        using RANSAC. Defaults to ``False``.
+    max_chunk_size : {int, None}, optional
+        The maximum number of channels to predict at once during channel-wise
+        RANSAC. If ``None``, RANSAC will use the largest chunk size that will
+        fit into the available RAM, which may slow down other programs on the
+        host system. If using window-wise RANSAC (the default) or not using
+        RANSAC at all, this parameter has no effect. Defaults to ``None``.
     random_state : {int, None, np.random.RandomState}, optional
         The random seed at which to initialize the class. If random_state is
         an int, it will be used as a seed for RandomState.
@@ -99,6 +112,8 @@ class PrepPipeline:
         prep_params,
         montage,
         ransac=True,
+        channel_wise=False,
+        max_chunk_size=None,
         random_state=None,
         filter_kwargs=None,
         matlab_strict=False,
@@ -133,7 +148,11 @@ class PrepPipeline:
         if self.prep_params["reref_chs"] == "eeg":
             self.prep_params["reref_chs"] = self.ch_names_eeg
         self.sfreq = self.raw_eeg.info["sfreq"]
-        self.ransac = ransac
+        self.ransac_settings = {
+            'ransac': ransac,
+            'channel_wise': channel_wise,
+            'max_chunk_size': max_chunk_size
+        }
         self.random_state = check_random_state(random_state)
         self.filter_kwargs = filter_kwargs
         self.matlab_strict = matlab_strict
@@ -189,9 +208,9 @@ class PrepPipeline:
         reference = Reference(
             self.raw_eeg,
             self.prep_params,
-            ransac=self.ransac,
             random_state=self.random_state,
-            matlab_strict=self.matlab_strict
+            matlab_strict=self.matlab_strict,
+            **self.ransac_settings
         )
         reference.perform_reference()
         self.raw_eeg = reference.raw
--- a/pyprep/prep_pipeline.py
+++ b/pyprep/prep_pipeline.py
@@ -43,7 +43,7 @@ class PrepPipeline:
         entire signal length ("channel-wise RANSAC", see `max_chunk_size`
         parameter). If ``False``, RANSAC will instead predict signals for all
         channels at once but over a number of smaller time windows instead of
-        over the entirem signal length ("window-wise RANSAC"). Channel-wise
+        over the entire signal length ("window-wise RANSAC"). Channel-wise
         RANSAC generally has higher RAM demands than window-wise RANSAC
         (especially if `max_chunk_size` is ``None``), but can be faster on
         systems with lots of RAM to spare. Has no effect if not using RANSAC.
--- a/pyprep/prep_pipeline.py
+++ b/pyprep/prep_pipeline.py
@@ -39,12 +39,15 @@ class PrepPipeline:
         Whether or not to use RANSAC for noisy channel detection in addition to
         the other methods in :class:`~pyprep.NoisyChannels`. Defaults to True.
     channel_wise : bool, optional
-        Whether RANSAC should predict signals for whole chunks of channels at
-        once instead of predicting signals for each RANSAC window
-        individually. Channel-wise RANSAC generally has higher RAM demands than
-        window-wise RANSAC (especially if `max_chunk_size` is ``None``), but can
-        be faster on systems with lots of RAM to spare.nHas no effect if not
-        using RANSAC. Defaults to ``False``.
+        Whether RANSAC should predict signals for chunks of channels over the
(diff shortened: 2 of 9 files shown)


a-hurst:
Whoops, accidentally submitted the PR before writing a description! Fixing that now.

sappelhoff:
gonna do a proper review later, but regarding the weird matprep correlation --> is that maybe because the data are already mean-centered? ... i.e., mean(x) is already around 0, so x - mean(x) is unnecessary and you can just square x directly? Just a hunch ... could be completely off.

a-hurst:
> gonna do a proper review later, but regarding the weird matprep correlation --> is that maybe because the data are already mean-centered? ... i.e., mean(x) is already around 0, so x - mean(x) is unnecessary and you can just square x directly? Just a hunch ... could be completely off.

That sounds like a good explanation. That said, it is generally going to be true that the real or predicted signal for each channel during any given 5-second window is centered on zero? I should probably compare the results directly with normal Pearson correlation to see how much of a difference it makes. Also, the "bad-by-correlation" code uses the normal numpy `corrcoef` correlation function and matches up identically with MATLAB PREP, so it's odd that they'd use this unique correlation method here but not there.

Regardless, if we ultimately decide that Pearson (or some other method of correlation) would make a better default then we'll probably need to adjust the RANSAC bad-correlation threshold as well, since the current one is tuned for the current calculation method. That makes me think this question might be better suited for a future issue/PR.

pyprep/ransac.py
```diff
@@ -127,7 +131,7 @@ def find_bad_by_ransac(
         random_ch_picks.append(picks)
 
     # Correlation windows setup
-    correlation_frames = corr_window_secs * sample_rate
+    correlation_frames = int(corr_window_secs * sample_rate)
```
sappelhoff, line 134: Why is this change necessary? If it *is* necessary, please also change the other parts in the code where `correlation_frames` is calculated - e.g., line 294 in find_noisy_channels.py

pyprep/utils.py
```diff
@@ -121,6 +144,37 @@ def _get_random_subset(x, size, rand_state):
     return sample
 
 
+def _correlate_arrays(a, b, axis=1):
```
sappelhoff, line 147: can we call this _mat_correlate_arrays? Or something the like? So that we better see that this comes from Matlab PREP?

pyprep/ransac.py
```diff
@@ -396,3 +411,126 @@ def _get_ransac_pred(
...
+    Returns
+    -------
+    predicted : np.ndarray
+        The median RANSAC-predicted EEG signal for the given window of data.
+
+    Notes
+    -----
+    In MATLAB PREP, the median signal is calculated by sorting the different
+    predictions for each EEG sample/channel from low to high and then taking the value
+    at the middle index (as calculated by `int(n_ransac_samples / 2.0)`) for each.
```
sappelhoff, line 518: in RST, inline codeblocks are rendered using double backticks:

```
# Markdown
`this is code`
```

```
# RST
``this is code``
```

Not to be confused with the situation where we mention function parameters in the docstring --> there we use single backticks :)

There are some more of these cases (see e.g., line 501)

pyprep/ransac.py
```diff
@@ -396,3 +411,126 @@ def _get_ransac_pred(
     interpol_mat = _make_interpolation_matrix(reconstr_pos, chn_pos)
     ransac_pred = np.matmul(interpol_mat, data[reconstr_picks, :])
     return ransac_pred
+
+
+def _ransac_correlations_alt(good_data, interpolation_mats, win_size, win_count):
```
sappelhoff, line 416: Can we give this a better name? Potentially even renaming the old "_ransac_correlations", so that we better see what each function does.

sappelhoff (commented):
Thanks Austin, could you perhaps adjust the PR title to be more descriptive of what is truly happening in terms of code? The current title

>  Make PyPREP's RANSAC produce identical results to MATLAB PREP

is ambiguous :-)

sappelhoff:
> That sounds like a good explanation. That said, it is generally going to be true that the real or predicted signal for each channel during any given 5-second window is centered on zero? 

I don't know, it was a shot in the dark.

> Regardless, if we ultimately decide that Pearson (or some other method of correlation) would make a better default then we'll probably need to adjust the RANSAC bad-correlation threshold as well, since the current one is tuned for the current calculation method. That makes me think this question might be better suited for a future issue/PR.

yes, good point :)

> Just tested directly, and it looks like the the results are pretty similar (although still somewhat divergent) between the correlation method MatPREP uses and the one we were using before (np.corrcoef).

that *does* suggest that mean(x) is indeed close to zero, and it makes little difference whether it's subtracted from x or not. 

I like `np.corrcoef` better though 🤔

a-hurst:
> I like `np.corrcoef` better though 🤔

Me too. I'm going to make the MatPREP correlation method a `matlab_strict` option in my new PR and leave `np.corrcoef` as the default.

pyprep/ransac.py
a-hurst: Good catch, thanks! Still learning the ropes with RST: my initial programming education wasn't big on documentation so I've had to gradually self-teach.

pyprep/utils.py
a-hurst: So far I've only used the `_mat` prefix for functions that mimic the behaviour of built-in MATLAB functions. I can add a `_matprep` prefix to differentiate, but I was thinking for my follow-up PR that I'd make `_correlate_arrays` like `_predict_median_signals` and add a `matlab_strict` flag to switch between PyPREP (`np.corrcoef`) and MatPREP behaviour.

pyprep/ransac.py
a-hurst: That was just so I could pass `correlation_frames` as an argument to `_ransac_correlations_alt` instead of `n`, since I thought it was more readable (it's used for array slicing, so it needs to be an int). I can change it elsewhere as well.

pyprep/utils.py
sappelhoff: >  but I was thinking for my follow-up PR that I'd make _correlate_arrays like _predict_median_signals and add a matlab_strict flag to switch between PyPREP (np.corrcoef) and MatPREP behaviour.

okay, as long as you don't forget about that, that sound better :+1:

pyprep/ransac.py
sappelhoff: We just need to be sure about the rounding logic here. So that with the new change we don't accidentally change a behavior that was working before. See below:

```
In [8]: np.arange(5.5)
Out[8]: array([0., 1., 2., 3., 4., 5.])

In [9]: np.arange(int(5.5))
Out[9]: array([0, 1, 2, 3, 4])
```

pyprep/utils.py
```diff
@@ -21,6 +21,29 @@ def _intersect(list1, list2):
     return list(set(list1).intersection(set(list2)))
 
 
+def _mat_round(x):
```
yjmantilla, line 24: I was thinking that maybe it was better using the whole "matlab" prefix instead of the "mat" one. I understand is more verbose but one could confuse "mat"  as being associated with a matrix.

In any case, this is just a minor suggestion, is not that relevant.

pyprep/utils.py
yjmantilla: I was thinking that maybe it was better using the whole "matlab" prefix instead of the "mat" one. I understand is more verbose but one could confuse "mat" as being associated with a matrix.

In any case, this is just a minor suggestion, is not that relevant.

yjmantilla (commented):
Looks good to me. Sorry for joining in late. I have been really busy with uni and work. Really looking forward to testing this implementation against a lot of files I processed through matlab's prep.

sappelhoff (commented):
@a-hurst are you taking a break from this PR to come back later? Or are there any blocking items that you need help with?

a-hurst:
> @a-hurst are you taking a break from this PR to come back later? Or are there any blocking items that you need help with?

Since I've broken this down into a bunch of separate PRs now, the only new thing this PR adds is support for is window-wise RANSAC, which is slower than the current method but has much lower RAM demands for longer recordings. To finish this up, I'm going to rebase on the current master branch and see if I can improve performance at all. I'll try and do the former today!

sappelhoff:
sounds great! --> but even if it's a bit slower, it'd still be a win for users with low RAM machines :+1:

a-hurst:
> sounds great! --> but even if it's a bit slower, it'd still be a win for users with low RAM machines 👍

When I pull this together, should I make window-wise the new default then, with channel-wise still as an option for people with high-RAM workstations who want to process as quickly as possible?

sappelhoff:
Sounds good!

a-hurst:
@sappelhoff Okay, so I think this is ready for review now: I haven't updated the `whats_new` yet and still need to finish exposing the public API for channel-wise RANSAC, but all the key parts are in place now and there are some things I wanted to get your thoughts on before finishing up.

The main thing I wanted to get feedback on is the repurposing of the `channel_wise` argument to `find_bad_by_ransac`: basically, I realized it would be confusing to have an argument for `window_wise` if that's going to be the default method, so I made `channel_wise` the argument to specify the channel-chunking method of RANSAC. I also added a new argument `max_chunk_size` to provide the same functionality as the old `channel_wise` argument (i.e. lets you run 1 channel per chunk) but with a bit more flexibility.

*However*, I realize that `channel_wise` is technically currently part of the public API and it's generally very bad practice to change the meaning of an argument between releases. My rationale here is that the current `channel_wise` functionality was never fully exposed (i.e., you couldn't select one-channel-at-a-time RANSAC from the `PrepPipeline` class, you could only use it if using `NoisyChannels` manually), and it was never formally introduced in the What's New, meaning it's unlikely to be used beyond the small circle of contributors here (@yjmantilla does this break anything for you?). Additionally, if I change the default value of `max_chunk_size` to 1 instead of none, `channel_wise` will behave exactly as it did before (though I think `None` is the more useful default).

Apart from that, my main changes are:

- Reworked channel-wise RANSAC to use the same pre-generated interpolation matrices as window-wise RANSAC, which speeds things up and reduces the amount of divergent/redundant code between methods
- Renamed the existing private channel-wise RANSAC functions to be consistent with the window-wise function names. Also rearranged their argument names/orders, variable names, and docstrings to match up more closely with their window-wise counterparts (for easier comparison/readability).
- Added a progress-printing function for window-wise RANSAC
- Overhauled the RANSAC unit tests considerably, including the addition of explicit comparison of the correlation matrices for window-wise/channel-wise for both regular and MATLAB-strict RANSAC to make sure window-wise and channel-wise produce identical results for the same seed.

Let me know what you think!

pyprep/ransac.py
a-hurst: Ah whoops, I forgot to do this in the latest round of changes. I'll push this fix along with any others you mention during the next round of review.

sappelhoff:
>  I realize that channel_wise is technically currently part of the public API and it's generally very bad practice to change the meaning of an argument between releases.

generally yes, but with the immense changes that are coming in version `0.4.0`, it's totally fine to break a couple of things if afterwards everything is better. Strict semantic versioning would make us increment a "Major" version even (so 0.3.1 -> 1.0.0, instead of 0.3.1 -> 0.4.0), but a 1.0.0 release would be too much of a statement at this point, I think.

> and it was never formally introduced in the What's New, meaning it's unlikely to be used beyond the small circle of contributors here

another good argument why breaking would be fine here

> though I think None is the more useful default

+1 to None :+1: 


would be interested in your opinion as well here @yjmantilla

pyprep/utils.py
```diff
@@ -362,6 +362,44 @@ def split_list(mylist, chunk_size):
     ]
 
 
+def print_progress(current, end, start=None, stepsize=1, every=0.1):
```
sappelhoff, line 365: why did you opt to code up you own progress tracking function instead of using [`tqdm`](https://github.com/tqdm/tqdm)? To avoid an additional dependency?

pyprep/find_noisy_channels.py
```diff
@@ -447,10 +448,18 @@ def find_bad_by_ransac(
...
-            Whether RANSAC should be performed one channel at a time (lower RAM
-            demands) or in chunks of as many channels as can fit into the
-            currently available RAM (faster). Defaults to ``False`` (i.e., using
-            the faster method).
+            Whether RANSAC should predict signals for whole chunks of channels
+            at once instead of predicting signals for each RANSAC window
+            individually. Channel-wise RANSAC generally has higher RAM demands
+            than window-wise RANSAC (especially if `max_chunk_size` is
+            ``None``), but can be faster on systems with lots of RAM to spare.
+            Defaults to ``False``.
```
sappelhoff, line 456: Okay I think I am a little confused by this.

We can run RANSAC in the following ways:

1. over all channels and over the whole time --> this is how I coded it back then
    1. needs huge amounts of RAM
    2. but if that RAM is available, it's fast
2. one channel at a time over the whole time --> this is what @yjmantilla implemented (if I recall correctly)
    1. much less RAM needed
    2. significantly slower
3. calculate RANSAC over all channels but over windows
    1. this is what MATLAB PREP does, and what autoreject does
    2. implemented by @a-hurst 
    3. Probably the best default

There would be a fourth potential option -> running single channels over windows ... so that'd be the slowest possible, but the lowest RAM ever (can probably even run on a raspberry pi :rofl: ).

Overall I am not sure how many of these options we should offer and how to document them.

Right now, I think one option should be fine: the matlab/autoreject way of doing all channels at once but over windows ... and if too little RAM is there, we just need to make the windows smaller ... right? (or would that affect the accuracy at some point?)

what are your opinions @a-hurst @yjmantilla

pyprep/utils.py
sappelhoff: It'd be fun to see how fast/slow your function is compared to what tqdm says:

> Overhead is low -- about 60ns per iteration (80ns with tqdm.gui), and is unit tested against performance regression. By comparison, the well-established ProgressBar has an 800ns/iter overhead.

pyprep/find_noisy_channels.py
a-hurst: I'd thought about that too (chunking-by-channel in window-wise RANSAC), but wasn't sure how to best merge the chunking code with the windowing stuff without a massive rewrite.

Right now, window-wise still takes less RAM than one-at-a-time channel-wise unless (channel count x 5 seconds) is larger than the length of the whole recording (i.e., one-channel-at-a-time only takes less RAM if a 32-channel recording is under 2.6 minutes, a 64-channel is under 5.3 mins, a 128-channel is under 10.6 mins, etc.). Thus, in most cases window-wise should be good enough on the RAM front. Still, I wouldn't be opposed to getting that down further: with enough optimization, I could probably get PyPREP running (slowly) without swapping on my old iMac G4!

pyprep/utils.py
a-hurst: Huh. My guess is that their code is probably better optimized but that my thing is still faster than ProgressBar (which I also didn't know existed until just now): I made no attempt to optimize my own approach, but I don't think there are any particularly expensive or wasteful function calls I'm doing per loop and I'm not trying to write as much to the terminal on every update.

a-hurst:
@sappelhoff Just thinking of how to implement the public API for this now: do you think these kinds of settings should be part of the `prep_params` dict, or should they be separate arguments on their own?

sappelhoff:
Probably separate parameters :thinking: and yes, the find_all_bads method should have these params as well.

pyprep/find_noisy_channels.py
yjmantilla: I think window-wise is better. I think remember talking bout this in another thread. In the usual use-case window-wise will usually consume less ram than the approach i took.

I think it would be nice to offer low-ram version even though it is the slowest, maybe fitting the largest chunk of channels could be done automatically like it is done currently in the channel wise version. but is not a priority. All the work @a-hurst  is doing is quite amazing, priority should be that matlab prep comparison.

So +1 making the window-wise the default

yjmantilla:
> @yjmantilla does this break anything for you?

It does not, and even if it would I would do a fork.

a-hurst:
@sappelhoff I've now added a proper public API for `channel_wise` and `max_chunk_size` and updated the `whats_new.rst`. I think this is ready for final review!

I'm still open to adding chunking support for window-wise as you suggested, but I think that might be better to do in a future PR given the amount of change in this one as-is: from a cursory look at what's involved I'd have to refactor both channel-wise and window-wise RANSAC a fair bit to make that work (see: all the wrapping try/catching re: RAM-checking that's currently not used at all for window-wise after the initial check further up). While working on this PR I had some thoughts on how to clean that up too, but again I didn't want to overload this with too many peripheral changes.

pyprep/find_noisy_channels.py
a-hurst: Thanks @yjmantilla! As for the MATLAB PREP comparison, I just got the [MATLAB side of that working today](https://github.com/a-hurst/matprep_artifacts) (automated MatPREP on GitHub actions, saving relevant data as artifacts for later testing) so expect some preliminary stuff on that end shortly :)

sappelhoff: image


:rofl:

pyprep/find_noisy_channels.py
```diff
@@ -144,22 +144,42 @@ def get_bads(self, verbose=False):
...
         """Call all the functions to detect bad channels.
 
         This function calls all the bad-channel detecting functions.
 
         Parameters
         ----------
-        ransac : bool
-            To detect channels by ransac or not.
+        ransac : bool, optional
+            Whether RANSAC should be for bad channel detection, in addition to
```
sappelhoff, line 155:
```suggestion
            Whether RANSAC should be used for bad channel detection, in addition to
```

pyprep/find_noisy_channels.py
```diff
@@ -144,22 +144,42 @@ def get_bads(self, verbose=False):
...
+            the other methods. RANSAC can detect bad channels that other methods
+            are unable to catch, but also slows down noisy channel detection
+            considerably. Defaults to ``True``.
+        channel_wise : bool, optional
+            Whether RANSAC should predict signals for whole chunks of channels
+            at once instead of predicting signals for each RANSAC window
+            individually. Channel-wise RANSAC generally has higher RAM demands
```
sappelhoff, line 162:
```suggestion
            Whether RANSAC should predict signals for chunks of channels over the entire
            signal length ("channel-wise RANSAC", see `max_chunk_size` parameter).
            If ``False``, RANSAC will instead predict signals for all channels at once
            but over a number of smaller time windows instead of over the entire
            signal length ("window-wise RANSAC").
            Channel-wise RANSAC generally has higher RAM demands
```

pyprep/find_noisy_channels.py
sappelhoff: please adjust this docstr according to https://github.com/sappelhoff/pyprep/pull/66/files#r624438768 for consistency :)

pyprep/prep_pipeline.py
```diff
@@ -38,6 +38,19 @@ class PrepPipeline:
     ransac : bool, optional
         Whether or not to use RANSAC for noisy channel detection in addition to
         the other methods in :class:`~pyprep.NoisyChannels`. Defaults to True.
+    channel_wise : bool, optional
```
sappelhoff, line 41: see https://github.com/sappelhoff/pyprep/pull/66/files#r624438768

pyprep/reference.py
```diff
@@ -32,6 +32,19 @@ class Reference:
     ransac : bool, optional
         Whether or not to use RANSAC for noisy channel detection in addition to
         the other methods in :class:`~pyprep.NoisyChannels`. Defaults to True.
+    channel_wise : bool, optional
```
sappelhoff, line 35: once more, see: https://github.com/sappelhoff/pyprep/pull/66/files#r624438768

pyprep/ransac.py
```diff
@@ -70,10 +72,17 @@ def find_bad_by_ransac(
         The duration (in seconds) of each RANSAC correlation window. Defaults to
         5 seconds.
     channel_wise : bool, optional
-        Whether RANSAC should be performed one channel at a time (lower RAM
```
sappelhoff, line 73: https://github.com/sappelhoff/pyprep/pull/66/files#r624438768

sappelhoff (approved):
Cool! I have only one major comment for the docstr, the rest looks good to go!

pyprep/find_noisy_channels.py
a-hurst: lol that commit was after I'd spent a good 30-40 minutes trying to figure out why `pop_saveset` was cryptically telling me I didn't have write access to the artifacts folder.

Turns out, if you use any keyword arguments in a MATLAB function call (e.g. `'savemode', 'onefile'`), that somehow means that positional arguments beyond the first one _no longer work_ and you have to use keyword arguments for everything else (i.e. `pop_saveset(EEG, outpath, 'savemode', 'onefile');` doesn't work, even though `pop_saveset(EEG, 'filename', outpath, 'savemode', 'onefile');` does and `pop_saveset(EEG,  outpath);` does too). And I thought passing argument names as strings was bad enough...

pyprep/find_noisy_channels.py
a-hurst: Nice and clear, I like it!

a-hurst:
> Cool! I have only one major comment for the docstr, the rest looks good to go!

Fixed in the latest commit, should be good to go now 🙂

pyprep/find_noisy_channels.py
```diff
@@ -144,22 +144,45 @@ def get_bads(self, verbose=False):
...
+            Whether RANSAC should be used for bad channel detection, in addition
+            to the other methods. RANSAC can detect bad channels that other
+            methods are unable to catch, but also slows down noisy channel
+            detection considerably. Defaults to ``True``.
+        channel_wise : bool, optional
+            Whether RANSAC should predict signals for chunks of channels over the
+            entire signal length ("channel-wise RANSAC", see `max_chunk_size`
+            parameter). If ``False``, RANSAC will instead predict signals for all
+            channels at once but over a number of smaller time windows instead of
+            over the entirem signal length ("window-wise RANSAC"). Channel-wise
```
sappelhoff, line 164:
```suggestion
            over the entire signal length ("window-wise RANSAC"). Channel-wise
```

pyprep/find_noisy_channels.py
```diff
@@ -447,10 +471,20 @@ def find_bad_by_ransac(
...
         channel_wise : bool, optional
-            Whether RANSAC should be performed one channel at a time (lower RAM
-            demands) or in chunks of as many channels as can fit into the
-            currently available RAM (faster). Defaults to ``False`` (i.e., using
-            the faster method).
+            Whether RANSAC should predict signals for chunks of channels over the
+            entire signal length ("channel-wise RANSAC", see `max_chunk_size`
+            parameter). If ``False``, RANSAC will instead predict signals for all
+            channels at once but over a number of smaller time windows instead of
+            over the entirem signal length ("window-wise RANSAC"). Channel-wise
```
sappelhoff, line 478:
```suggestion
            over the entire signal length ("window-wise RANSAC"). Channel-wise
```

pyprep/prep_pipeline.py
```diff
@@ -38,6 +38,22 @@ class PrepPipeline:
...
+        Whether RANSAC should predict signals for chunks of channels over the
+        entire signal length ("channel-wise RANSAC", see `max_chunk_size`
+        parameter). If ``False``, RANSAC will instead predict signals for all
+        channels at once but over a number of smaller time windows instead of
+        over the entirem signal length ("window-wise RANSAC"). Channel-wise
```
sappelhoff, line 46:
```suggestion
        over the entire signal length ("window-wise RANSAC"). Channel-wise
```

pyprep/ransac.py
```diff
@@ -70,10 +72,20 @@ def find_bad_by_ransac(
...
-        demands) or in chunks of as many channels as can fit into the currently
-        available RAM (faster). Defaults to ``False`` (i.e., using the faster
-        method).
+        Whether RANSAC should predict signals for chunks of channels over the
+        entire signal length ("channel-wise RANSAC", see `max_chunk_size`
+        parameter). If ``False``, RANSAC will instead predict signals for all
+        channels at once but over a number of smaller time windows instead of
+        over the entirem signal length ("window-wise RANSAC"). Channel-wise
```
sappelhoff, line 79:
```suggestion
        over the entire signal length ("window-wise RANSAC"). Channel-wise
```

pyprep/reference.py
```diff
@@ -32,6 +32,19 @@ class Reference:
...
+        Whether RANSAC should predict signals for whole chunks of channels at
+        once instead of predicting signals for each RANSAC window
+        individually. Channel-wise RANSAC generally has higher RAM demands than
+        window-wise RANSAC (especially if `max_chunk_size` is ``None``), but can
+        be faster on systems with lots of RAM to spare.nHas no effect if not
+        using RANSAC. Defaults to ``False``.
```
sappelhoff, line 41:
```suggestion
        Whether RANSAC should predict signals for chunks of channels over the
        entire signal length ("channel-wise RANSAC", see `max_chunk_size`
        parameter). If ``False``, RANSAC will instead predict signals for all
        channels at once but over a number of smaller time windows instead of
        over the entire signal length ("window-wise RANSAC"). Channel-wise
        RANSAC generally has higher RAM demands than window-wise RANSAC
        (especially if `max_chunk_size` is ``None``), but can be faster on
        systems with lots of RAM to spare. Has no effect if not using RANSAC.
        Defaults to ``False``.
```

sappelhoff (approved):
you had a tiny typo and I think you forgot to update one of the many docstrs. These suggestions should fix it. I'll accept them and then merge -> but feel free to double check :-)


Thanks a lot! One leap closer to 0.4 :clap:

PR #71, opened 2021-04-26
Make removeTrend output identical to MATLAB PREP

### LINKED ISSUE #55: removeTrend produces different values in PyPREP than MATLAB PREP
Splitting this off from #49 to keep things organized. Basically, the removeTrend function (which essentially just runs a highpass filter to remove frequencies below 1 Hz) produces different values in PyPREP's implementation (by default, a wrapper around `mne.filter.filter_data`) than in MATLAB PREP's implementation (by default, a wrapper around EEGLAB's `pop_eegfiltnew`).

For comparison I loaded the same .set file into both EEGLAB and MNE, applied `removeTrend` to both, then re-saved the .set from EEGLAB so I could load it into MNE and compare. The PSD plots for both look identical to me, but the numeric values themselves via `raw._data` look quite different:

<details>
  <summary>Comparisons of MATLAB and PyPREP values (click to unhide)</summary>

First sample of first 10 channels in MATLAB PREP before trend removal:
```
signal.data(:, 1)

ans =

   1.0e+04 *

   -0.3335
    1.0037
   -0.3581
   -0.1795
    0.9913
   -0.7414
   -0.7198
   -0.5665
    0.5557
   -1.2237
```

 First sample of first 10 channels in MATLAB PREP after trend removal:
```
signal2.data(:, 1)

ans =

   -3.7068
   -4.8012
   -0.8008
   -2.5809
   -1.5205
    0.1093
    1.9928
   -4.6906
  -10.2859
    1.6814
```

First sample of first 10 channels in PyPREP before trend removal:
```
raw_copy._data[:10, 0]
array([-0.00333541,  0.01003745, -0.00358109, -0.00179516,  0.00991251,
       -0.00741443, -0.00719762, -0.00566537,  0.00555661, -0.01223686])
```

First sample of first 10 channels in PyPREP after trend removal:
```
raw_filtered._data[:10, 0]
array([-4.11658012e-19, -1.79973325e-18,  6.98802181e-19,  2.89261751e-19,
        2.50128829e-18, -9.28348110e-19,  7.57247455e-19,  3.55753838e-19,
        3.82858892e-19, -7.01343280e-19])
```
</details>

Since the plots look the same the differences might not practically matter all that much, but if we want to be able to do formal (or informal) 1:1 comparisons of PyPREP results to MatPREP, having identical results from this function would be helpful, especially since this always happens before any of the NoisyChannels operations.

a-hurst:
# PR Description

Closes #55.

Basically, MNE and EEGLAB calculate FIR filter parameters slightly differently (specifically, the length of the filter). To bypass this, I've added a new function that calculates the parameter values the same way that EEGLAB does and uses SciPy's `signal.firwin` to turn them into a MatPREP-identical filter. Then, I use the same internal method MNE's `filter.filter_data` does to apply the filter to the signal, producing identical output to MatPREP's `removeTrend`.

# Merge Checklist

### DIFF
--- a/docs/matlab_differences.rst
+++ b/docs/matlab_differences.rst
@@ -18,6 +18,31 @@ internal math.
     :depth: 3
 
 
+Differences in Signal Detrending
+--------------------------------
+
+In the PREP pipeline, trends (i.e., slow drifts in EEG baseline signal) are
+temporarily removed from the data prior to line-noise removal with CleanLine
+as well as prior to bad channel detection via :class:`~pyprep.NoisyChannels`,
+which occurs at multiple points during robust re-referencing. This is done to
+improve the accuracy of both of these processes, which are sensitive to
+influence from trends in the signal.
+
+In MATLAB PREP, the default method of trend removal is to use EEGLAB's
+``pop_eegfiltnew``, which creates and applies an FIR high-pass filter to the
+data. MNE's :func:`mne.filter.filter_data` offers similar functionality, but
+uses slightly different filter creation math and a different filtering
+algorithm such that its results and subsequent :class:`~pyprep.NoisyChannels`
+values also differ slightly (on the order of ~0.002) for RANSAC correlations.
+
+Because the practical differences are small and MNE's filtering is fast and
+well-tested, PyPREP defaults to using :func:`mne.filter.filter_data` for
+high-pass trend removal. However, for exact numerical compatibility, PyPREP
+has a basic re-implementaion of EEGLAB's ``pop_eegfiltnew`` in Python that
+produces identical results to MATLAB PREP's ``removeTrend`` when
+``matlab_strict`` is set to ``True``.
+
+
 Differences in RANSAC
 ---------------------
 
--- a/docs/matlab_differences.rst
+++ b/docs/matlab_differences.rst
@@ -22,7 +22,7 @@ Differences in Signal Detrending
 --------------------------------
 
 In the PREP pipeline, trends (i.e., slow drifts in EEG baseline signal) are
-temporarily removed from the data prior to line-noise removal with CleanLine
+temporarily removed from the data prior to adaptive line-noise removal
 as well as prior to bad channel detection via :class:`~pyprep.NoisyChannels`,
 which occurs at multiple points during robust re-referencing. This is done to
 improve the accuracy of both of these processes, which are sensitive to
--- a/pyprep/removeTrend.py
+++ b/pyprep/removeTrend.py
@@ -48,8 +48,12 @@ def removeTrend(
     :func:``mne.filter.filter_data`` unless `matlab_strict` is ``True``, in
     which case it is performed using a minimal re-implementation of EEGLAB's
     ``pop_eegfiltnew``. Local detrending is performed using a Python
-    re-implementation of the ``runline`` command from the chronux_2 MATLAB
-    package.
+    re-implementation of the ``runline`` function from the Chronux package for
+    MATLAB [1]_.
+
+    References
+    ----------
+    .. [1] http://chronux.org/
 
     """
     if len(EEG.shape) == 1:
@@ -114,21 +118,28 @@ def removeTrend(
 
 
 def runline(y, n, dn):
-    """Implement chronux_2 runline command for performing local linear regression.
+    """Perform local linear regression on a channel of EEG data.
+
+    A re-implementation of the ``runline`` function from the Chronux package
+    for MATLAB [1]_.
 
     Parameters
     ----------
     y : np.ndarray
-        Input from one EEG channel.
+        A 1-D array of data from a single EEG channel.
     n : int
-        length of the detrending window.
+        Length of the detrending window.
     dn : int
-        length of the window step size.
+        Length of the window step size.
 
     Returns
     -------
     y: np.ndarray
-       Detrended EEG signal for one channel.
+       The detrended signal for the given EEG channel.
+
+    References
+    ----------
+    .. [1] http://chronux.org/
 
     """
     nt = y.shape[0]
--- a/pyprep/removeTrend.py
+++ b/pyprep/removeTrend.py
@@ -13,6 +13,7 @@ def removeTrend(
     detrendType="high pass",
     detrendCutoff=1.0,
     detrendChannels=None,
+    matlab_strict=False,
 ):
     """Remove trends (i.e., slow drifts in baseline) from an array of EEG data.
 
@@ -31,6 +32,10 @@ def removeTrend(
     detrendChannels : {list, None}, optional
         List of the indices of all channels that require detrending/filtering.
         If ``None``, all channels are used (default).
+    matlab_strict : bool, optional
+        Whether or not detrending should strictly follow MATLAB PREP's internal
+        math, ignoring any improvements made in PyPREP over the original code
+        (see :ref:`matlab-diffs` for more details). Defaults to ``False``.
 
     Returns
     -------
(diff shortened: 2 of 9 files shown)


a-hurst:
Whoops, looks like I broke something by accident. Serves me right for being too lazy to run unit tests locally before pushing.

a-hurst:
Okay, so I may have jumped the gun a bit: I was looking at the wrong values in MATLAB (the pre-filtered ones) and my FIR filter was also broken (missing the big spike in the middle), so with the two issues combined it looked like it was working.

I've fixed the custom filter itself so it's *actually* identical to MATLAB's now, but MNE's filter applying function still produces values from EEGLAB's so I'm giving this a rest for now. The actual filter application code in MATLAB looks reasonably simple, so maybe on a future day I'll try adapting that too.

sappelhoff:
If something is unclear from the MNE side, you can always pose a question on https://mne.discourse.group :-)

a-hurst:
@sappelhoff So it absolutely wasn't worth the total time and effort, but sunk cost fallacy kept me going until the end: in the latest commit I've implemented the equivalent of MATLAB's `firfilt` in Python using SciPy/NumPy.

Now NoisyChannels gives me identical RANSAC correlations regardless of whether I use a .set that's been pre-highpassed in MATLAB (what I've been doing so far for tests) or I do the high-passing in PyPREP with removeTrend. On the flip side, this method is also about 4x slower than the previous one (~8 seconds vs. ~2 seconds for a 64-channel, 694016-sample file), and I really don't know much about filter theory or signal processing so MNE's filtering could be technically superior at trend-removal than the code here.

My current thinking is that maybe this should just be wrapped in `matlab_strict` unless there's also a technical reason to prefer this style of filtering over the old one? Below I've included a comparison of the RANSAC correlations before and after this PR to illustrate how much of a difference this all makes. Let me know what you think!

<details>
  <summary>Comparisons of RANSAC correlations between removeTrend methods (click to unhide)</summary>

Current removeTrend method (MNE's `filter_data`):
```
array([[0.89438387, 0.82456243, 0.91580898, ..., 0.98728151, 0.95233041,
        0.94015163],
       [0.91034277, 0.9600703 , 0.64722385, ..., 0.97487885, 0.83335142,
        0.97426197],
       [0.81404095, 0.80873159, 0.80359327, ..., 0.96677663, 0.95435911,
        0.88024998],
       ...,
       [0.8676419 , 0.89621802, 0.8646553 , ..., 0.99084073, 0.96557544,
        0.98570571],
       [0.98061229, 0.95732037, 0.93206238, ..., 0.99242296, 0.97696497,
        0.9704131 ],
       [0.95575104, 0.84842815, 0.7716721 , ..., 0.99250605, 0.95400638,
        0.95062305]])
```

`_eeglab_create_highpass` + `_eeglab_fir_filter` (same as values when using pre-highpassed .set from MatPREP & commenting out PyPREP's `removeTrend`):
```
array([[0.89682968, 0.82714556, 0.91665897, ..., 0.98672507, 0.95110975,
        0.94141531],
       [0.91036137, 0.96005206, 0.64718089, ..., 0.97489243, 0.8333695 ,
        0.97430253],
       [0.81397999, 0.80860207, 0.80386369, ..., 0.96684684, 0.95454648,
        0.8802559 ],
       ...,
       [0.867604  , 0.89619309, 0.86480001, ..., 0.99082554, 0.9655723 ,
        0.98570184],
       [0.98061141, 0.95732534, 0.93203051, ..., 0.99242111, 0.97695023,
        0.97039977],
       [0.95548201, 0.8487357 , 0.77087238, ..., 0.99259436, 0.95389967,
        0.95089473]])
```

</details>

sappelhoff:
> My current thinking is that maybe this should just be wrapped in matlab_strict

yes! Let's do that and keep the mne detrending as a default :+1: 

the differences are quite tiny but it's good you got it to align perfectly --> yet as you say, we don't know how "good" and/or "robust" your code is ... at least relative to MNE's filtering implementation that is tested every minute by lots of users.

a-hurst:
If you know anyone who knows their stuff re: filtering and signal processing theory, it'd be interesting to hear their take on the functional difference between the EEGLAB and MNE methods. Since we're wrapping this in `matlab_strict` (and thus making it another "deliberate difference"), it'd be nice to know & be able to explain on the 'differences' RST page if MNE's approach is technically/theoretically better than EEGLAB's (beyond the obvious maintenance/speed concerns we've already discussed).

a-hurst:
@sappelhoff Okay, think I've got this sorted now! In addition to wrapping this in `matlab_strict` and documenting the difference in the deliberate differences page, I also cleaned up the docstrings for `removeTrend` a bit and changed `sample_rate` from an optional argument with a default to a required one since that seemed to make more sense.

If either of us learn any more about the costs/benefits of MNE's filtering algorithm vs. EEGLAB's, I'll update the deliberate differences section accordingly.

docs/matlab_differences.rst
```diff
@@ -18,6 +18,31 @@ internal math.
     :depth: 3
 
 
+Differences in Signal Detrending
+--------------------------------
+
+In the PREP pipeline, trends (i.e., slow drifts in EEG baseline signal) are
+temporarily removed from the data prior to line-noise removal with CleanLine
```
sappelhoff, line 25: can we get a link to CleanLine here? or some further info in brackets about it?

pyprep/removeTrend.py
```diff
@@ -4,51 +4,72 @@
...
     Notes
     -----
-    Filtering is implemented using the MNE filter function mne.filter.filter_data.
-    Local detrending is the python implementation of the chronux_2 runline command.
+    High-pass filtering is implemented using the MNE filter function
+    :func:``mne.filter.filter_data`` unless `matlab_strict` is ``True``, in
+    which case it is performed using a minimal re-implementation of EEGLAB's
+    ``pop_eegfiltnew``. Local detrending is performed using a Python
+    re-implementation of the ``runline`` command from the chronux_2 MATLAB
+    package.
```
sappelhoff, line 52: ideally you would add a reference here something like

```suggestion
    re-implementation of the ``runline`` command from the chronux_2 MATLAB
    package [1]_.
    
    References
    ----------
    .. [1] https:// .... link to chronux 2 ----
```

pyprep/utils.py
```diff
@@ -108,6 +109,103 @@ def _mat_iqr(arr, axis=None):
...
+    Because MNE's ``mne.filter.filter_data`` and EEGLAB's ``pop_eegfiltnew``
+    calculate filter parameters slightly differently, this function is
+    used to precisely match EEGLAB & MATLAB PREP's filtering method.
+
+    """
+    TRANSITION_WIDTH_RATIO = 0.25
+    trans_bandwidth = cutoff if cutoff < 2 else cutoff * TRANSITION_WIDTH_RATIO
+
+    # Calculate parameters for constructing filter
+    order = 3.3 / (trans_bandwidth / srate)
```
sappelhoff, line 140: where does the 3.3 come from? for magic numbers like these it's sometimes good style to do:

```
default_head_circumference_in_mm = 89
my_thing = default_head_circumference_in_mm * funky_tuple
```

instead of


```
my_thing = 89 * funky_tuple
```

sappelhoff (approved):
lgtm, only some tiny clarifications needed if possible

docs/matlab_differences.rst
a-hurst: Sure, I'll add a link to the relevant MNE function.

pyprep/removeTrend.py
a-hurst: Will do! The original docstring didn't have a reference, but now's as good a time as any to add one.

pyprep/utils.py
a-hurst: I honestly have no clue what the 3.3 is supposed to represent, it's not defined clearly in the EEGLAB code and my Google results haven't turned up a proper name for what it's supposed to be doing. The best I've found is Table 4.3 (page 7) in [this PDF](https://www.staff.ncl.ac.uk/oliver.hinton/eee305/Chapter4.pdf), which gives 3.3 as the unnamed constant to use when calculating the transition bandwidth for a Hamming window (here we're reversing that equation to get the filter order based on transition bandwidth & sample rate).

Does `HAMMING_CONSTANT` (or something to that effect) work for you?

sappelhoff: Ok, interesting. Yes that'd work - preferably with a comment that it's unclear

docs/matlab_differences.rst
a-hurst: I decided to remove the reference to CleanLine and just put "adaptive line-noise removal". Since the relevant MNE function defaults to a different sort of filter, adding a link and explaining the correct args was just going to make that sentence unwieldly. Hopefully that works!

sappelhoff (approved):
thanks a bunch!

a-hurst:
@sappelhoff So I'm currently taking a grad course on signal processing, and I think I finally understand the key difference between MNE's highpass filter and EEGLAB's: MNE defaults to applying the filter using a zero-phase method that doesn't result in any phase shift in the signal (the equivalent of `filtfilt` in MATLAB/Scipy), whereas the EEGLAB filter function I re-implemented here only does a single forward pass (using `lfilter`), so you end up with phase shift in the signal.

As far as I can tell, other than theoretically being faster (which it isn't in the current implementation) there's absolutely no reason to use a forwards-only filter like EEGLAB does instead of the backwards-forwards zero-phase approach used by MNE. I'm guessing this is a weird legacy carry-over for EEGLAB from the days when people were running EEG analyses on Pentium 3's and PowerPC G4's so CPU cycles were at a major premium. I'll update the "differences" docs to reflect this when I get a chance!

sappelhoff:
>  whereas the EEGLAB filter function I re-implemented here only does a single forward pass (using lfilter), so you end up with phase shift in the signal.

:scream: gosh ...

> I'm guessing this is a weird legacy carry-over for EEGLAB from the days when people were running EEG analyses on Pentium 3's and PowerPC G4's so CPU cycles were at a major premium

you are probably right, EEGLAB has been around for quite some time!

great that you remembered this issue!

Issue #74, opened 2021-05-04
RANSAC random seeds work differently in PyPREP than in MATLAB PREP


a-hurst:
In MATLAB PREP, the random seed for RANSAC during `findNoisyChannels` is always set to the same value (435656) immediately before RANSAC is run. This means that, given the same set of 'good' channels, two consecutive runs of RANSAC during re-referencing will have identical channel picks.

In PyPREP, the random seed is set at the onset of `PrepPipeline` or `Reference` and gets made into a Numpy RandomState object. This means that the random state is going to be different for every RANSAC run during re-referencing, resulting in different re-referencing results than MATLAB PREP and potentially higher variability between iterations (resulting in more channels being flagged as bad). Actually, given how random subsets are calculated in PyPREP/MatPREP (generating random floats from 0 to 1, multiplying them by the number of channels to choose from, and then popping the channel with the nearest index repeatedly in a loop), using the same seed across runs would mean that *similar* channels would get chosen across RANSAC runs, assuming that adjacent channels in the data are often also going to be adjacent spatially.

This can be easily fixed with a PR, but before I do I just want to double-check whether this is a behaviour we want to match? I mean, MATLAB's approach would lead to less variability in NoisyChannels results during re-referencing, but it would also mean that RANSAC's choices aren't fully random and would probably cause problems in an instance where a random seed makes really bad random picks the first time around.

What do you think, should this be default behaviour or another one for `matlab_strict`?

sappelhoff:
very interesting ... thanks for looking into that! :)

Intuitively, the pyprep approach sounds more sensible to me ... 

>  This means that the random state is going to be different for every RANSAC run during re-referencing, resulting in different re-referencing results than MATLAB PREP and potentially higher variability between iterations (**resulting in more channels being flagged as bad)**

I **made the important part bold**. --> The big question is whether the additional channels (compared to matprep) that are flagged bad, are sensibly flagged bad ... or whether it's a bad flagging.

That's something to check with visual inspection on 3 or 4 separate datasets I guess. If the finding is: pyprep does flag more channels but it looks sensible, then I'd say we should leave pyprep as is, and provide the matprep way in `matlab_strict` and document it in your nice doc page, Austin. Otherwise, we should make the matlab way the default.

a-hurst:
> That's something to check with visual inspection on 3 or 4 separate datasets I guess.

[...]

PR #78, opened 2021-05-04
Change initial re-referencing state in Reference to match Matlab PREP

### LINKED ISSUE #75: First-pass average referencing differs between PyPREP and Matlab PREP
In MATLAB PREP, robust re-referencing starts by detecting noisy channels, calculating the median signal excluding all channels bad by NaN, bad by flat, or bad by low SNR, and then re-referencing the data using that signal. PyPREP does almost the same thing, except it leaves in channels bad by low SNR which results in numeric differences when those sorts of channels are present:

https://github.com/sappelhoff/pyprep/blob/051edd6e34360b9e6cdae39ede93a84cda895109/pyprep/reference.py#L233-L240

Based on the above code & comment, it looks like this difference is deliberate? Is there an intentional reason to include these channels when calculating a median reference? If so, I'll make it a `matlab_strict` thing and document it in the "deliberate differences" page. If not, I'll just make a PR to add those in.

Looking at the MatPREP Git history, it looks like the SNR channels were a deliberate addition to the list of exclusions from the initial reference, added when the default method of calculating the initial reference signal was changed from "huber mean" (?) to median. Maybe it's worth asking Dr. Robbins about?

Anyway, if I add "bad-by-SNR" channels to the list of initial unusables in PyPREP, the initial average-referenced signal is identical to the one in MATLAB PREP.

### LINKED ISSUE #76: PyPREP starts re-referencing with more 'bad' channels than Matlab PREP
In PyPREP, we do a first pass of NoisyChannels prior to initial average referencing to

a) keep track of what the original channels flagged as "bad" were, and

b) figure out which channels are "unusable" for median-referencing before we start the actual NoisyChannels -> interpolate bads -> re-reference loop.

In PyPREP, we use the bad channels from this initial pass of NoisyChannels as the starting state for the subsequent loop (and once a channel is considered 'bad', it's permanently excluded from being a reference channel during re-referencing). See the last line here, where `self.noisy_channels` gets initialized with a copy of `self.noisy_channels_original`:

https://github.com/sappelhoff/pyprep/blob/051edd6e34360b9e6cdae39ede93a84cda895109/pyprep/reference.py#L216-L230

MatPREP, however, only adds the 'unusable' channels (i.e. those bad by NaN, flat, or SNR) to the starting state of permanently bad channels, leaving the deviation, HF noise, correlation, dropout, and RANSAC initial bads empty.

Personally I think MatPREP's behaviour makes more sense here, given that the signal hasn't necessarily been average-referenced coming into PyPREP (e.g. could be referenced to Cz) and the initial average referencing could change the "bad" status for a lot of channels (e.g. if Cz was particularly noisy). In fact, I'd go one further and say that "bad-by-flat" channels from the first NoisyChannels pass shouldn't be permanently excluded either for this same reason: if it's flat because it's a reference channel, it won't be flat anymore after the initial average referencing.

a-hurst:
# PR Description

Closes #75 and closes #76. This PR fixes two separate (but related) issues:

- Unlike PyPREP, MatPREP considers bad-by-SNR channels to be 'unuseable' for its initial average reference (in addition to bad-by-NaN and bad-by-flat channels)
- Unlike PyPREP, MatPREP only flags the above "unusable" channels for permanent exclusion from being reference channels after the first noisy channels pass. PyPREP includes bad channels detected by all methods from this pass as permanently unusable, which doesn't necessarily make sense if the data isn't already average referenced (e.g. single-channel referenced or CMS/DRL).

With these fixes, PyPREP and Matlab PREP have identical values during re-referencing up until the second pass of NoisyChannels (see #74).

# Merge Checklist

### DIFF
--- a/docs/whats_new.rst
+++ b/docs/whats_new.rst
@@ -43,6 +43,8 @@ Changelog
 - Added a ``matlab_strict`` method for high-pass trend removal, exactly matching MATLAB PREP's values if ``matlab_strict`` is enabled, by `Austin Hurst`_ (:gh:`71`)
 - Added a window-wise implementaion of RANSAC and made it the default method, reducing the typical RAM demands of robust re-referencing considerably, by `Austin Hurst`_ (:gh:`66`)
 - Added `max_chunk_size` parameter for specifying the maximum chunk size to use for channel-wise RANSAC, allowing more control over PyPREP RAM usage, by `Austin Hurst`_ (:gh:`66`)
+- Changed :class:`~pyprep.Reference` to exclude "bad-by-SNR" channels from initial average referencing, matching MATLAB PREP behaviour, by `Austin Hurst`_ (:gh:`78`)
+- Changed :class:`~pyprep.Reference` to only flag "unusable" channels (bad by flat, NaNs, or low SNR) from the first pass of noisy detection for permanent exclusion from the reference signal, matching MATLAB PREP behaviour, by `Austin Hurst`_ (:gh:`78`)
 
 Bug
 ~~~
--- a/pyprep/reference.py
+++ b/pyprep/reference.py
@@ -230,15 +230,11 @@ class Reference:
         self.noisy_channels = self.noisy_channels_original.copy()
         logger.info("Bad channels: {}".format(self.noisy_channels))
 
+        # Determine channels to use/exclude from initial reference estimation
         self.unusable_channels = _union(
-            noisy_detector.bad_by_nan, noisy_detector.bad_by_flat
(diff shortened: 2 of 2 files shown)


a-hurst:
Looks like the Ubuntu check failed because it couldn't reach Physionet to download the test EDF. Is there any way to re-run without pushing another commit?

sappelhoff (approved):

sappelhoff:
Thanks a bunch!

Short info:

When you write

> Closes #75 and #76. 

then GH will only close the first one automatically once the PR is merged.

I edited this to:

> Closes #75 and closes #76.

PR #81, opened 2021-05-06
Fix handling of channels with dropouts (intermittent flat regions) within NoisyChannels

### LINKED ISSUE #80: Dropout channel detection broken in PyPREP
Working on #79, I modified `matlab_artifacts` to simulate a "dropout" channel (I set the second 25% and last 25% of the channel to 0). This channel is successfully flagged as a bad-by-dropout by MatPREP, but in PyPREP the addition of this channel breaks NoisyChannels completely, seriously messing up the correlation windows and leaving too few usable channels to use for RANSAC.

I've been looking into it for a bit and it seems to be due to differences in how NaNs are generated during correlation between Numpy and MATLAB, but haven't figured out the exact source yet.

a-hurst:
# PR Description

Closes #80. Because Numpy's quantile function handles NaN values differently than MATLAB, finding bad-by-correlation channels in `NoisyChannels` broke completely if there were any correlation windows where a channel had a flat signal. Also, the code that was meant to replace NaN correlation values with zeros was accidentally doing so in the wrong matrix. Both of these are fixed by this PR!

# Merge Checklist

### DIFF
--- a/pyprep/find_noisy_channels.py
+++ b/pyprep/find_noisy_channels.py
@@ -382,7 +382,7 @@ class NoisyChannels:
                     np.isnan(channel_correlation[i, j]) or np.isnan(noiselevels[i, j])
                 )
                 if drop[i, j] == 1:
-                    channel_deviations[i, j] = 0
+                    channel_correlation[i, j] = 0
                     noiselevels[i, j] = 0
         maximum_correlations[self.channels_interpolate, :] = np.transpose(
             channel_correlation
--- a/pyprep/find_noisy_channels.py
+++ b/pyprep/find_noisy_channels.py
@@ -366,15 +366,17 @@ class NoisyChannels:
         for k in range(0, w_correlation):
             eeg_portion = np.transpose(np.squeeze(EEG_new_win[:, :, k]))
             data_portion = np.transpose(np.squeeze(data_win[:, :, k]))
-            window_correlation = np.corrcoef(np.transpose(eeg_portion))
+            with np.errstate(invalid='ignore'):  # suppress divide-by-zero warnings
+                window_correlation = np.corrcoef(np.transpose(eeg_portion))
             abs_corr = np.abs(
                 np.subtract(window_correlation, np.diag(np.diag(window_correlation)))
             )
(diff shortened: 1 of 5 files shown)


a-hurst:
Ah whoops, I accidentally based this off the MATLAB comparison test branch. Rebasing on master now.

pyprep/find_noisy_channels.py
```diff
@@ -366,23 +366,25 @@ def find_bad_by_correlation(
         for k in range(0, w_correlation):
             eeg_portion = np.transpose(np.squeeze(EEG_new_win[:, :, k]))
             data_portion = np.transpose(np.squeeze(data_win[:, :, k]))
-            window_correlation = np.corrcoef(np.transpose(eeg_portion))
+            with np.errstate(invalid='ignore'):  # suppress divide-by-zero warnings
+                window_correlation = np.corrcoef(np.transpose(eeg_portion))
```
sappelhoff, line 370: in which cases would the `np.errstate(invalid='ignore')` be necessary? And could it results in `window_correlation` being NaN?

sappelhoff (commented):
Ah thanks, now I see what you meant before. You solved it well, I think. Just one comment.

pyprep/find_noisy_channels.py
a-hurst: Yep, inside `np.corrcoef` it uses the standard deviation of the inputs at some point in the calculation, and since for a dropout channel the SD is going to be 0, we get NaNs because of the division-by-zero error (and a corresponding RuntimeWarning message). However, the current code uses those NaNs to determine which channels are dropout channels for each window so things would need to be refactored a fair bit to prevent the NaNs in the first place.

I'm thinking of doing a proper refactor of NoisyChannels at some point once MATLAB comparison is merged, so I went with the path of least resistance for now.

a-hurst: To clarify, the `with np.errstate` is just to suppress some division-by-zero runtime warnings that are currently inevitable/expected based on how the code is structured.

sappelhoff: that makes sense, thanks for clarifying.

sappelhoff (approved):

Issue #83, opened 2021-05-10
PREP can't always flag low amplitude channels as bad-by-deviation


a-hurst:
I've been picking away at refactoring the `NoisyChannels` unit tests so that each 'bad-by' type is covered by a separate test. In the process, though, I think I've discovered a weak spot in the original PREP logic: if I use a different test file, the 'bad-by-low-deviation test' below stops working:

https://github.com/sappelhoff/pyprep/blob/bc46978b912737bc3b843438a3cbe1cd9d4dc33d/tests/test_find_noisy_channels.py#L51-L60

This holds true even if I replace the "divide by 10" with a "divide by 100,000": the "robust channel deviation" z-score seems to plateau at around -4, never reaching the +/- 5 threshold for being bad by deviation.

Looking at the actual math, this makes sense: 

1. PREP calculates the variance within each channel using 0.7413 * the interquartile range for the signal.
2. PREP calculates the median and variance (again, 0.7413 * the IQR) of those values and uses them to do a non-traditional Z-score of the channel variances (i.e., `(variance - median_variance) / variance_of_the_variance`).
3. PREP compares the absolute values of those Z-scores to a threshold (default = 5) and flags any channels that exceed it as "bad-by-deviation"

The problem here is that in step 2, the variances calculated in step 1 have a minimum value of zero. As such, even a channel with an amplitude of zero won't be detected as bad-by-deviation if the median isn't at least 5x the variance of the variances.

A quick-and-dirty fix is to z-score the *log* of the channel variances, which makes the variances more normally-distributed and thus makes the detection of low-amplitude channels much easier. However, this also increases the threshold for detecting high-amplitude bad-by-deviation channels, with the required multiplication factor going from 2.4 to 3.9 for my test channel.

Any ideas on how to better handle this?

sappelhoff:
Wouldn't channels with suspiciously like deviation be flagged by the bad by flat check? Else, would decreasing the threshold from 5 to 3.x help? (3.x being that canonical value that I always forget the meaning of and what x actually is)

a-hurst:
> I think the PREP devs didn't think of what they did as a statistical test that matches certain assumptions

Probably, but I think it still makes sense to treat it like one: we're testing to see whether each channel's amplitude meaningfully deviates from all the others, so it's still a statistical procedure (like outlier removal for reaction times) if not exactly a statistical test.

[...]

PR #84, opened 2021-05-10
Refactor of NoisyChannels tests and split into separate methods

a-hurst:
# PR Description

This PR takes the `test_findnoisychannels` function and splits it into separate unit tests for each type of bad channel, making it easier to maintain and figure out which specific part(s) are broken by a given change.

I also changed the test file from the current (very noisy) one to one with only two noisy channels to start with, saving time repeatedly running `NoisyChannels` and interpolating bads at the start to try and produce a clean recording for subsequent tests. This also has the advantage of not depending on `NoisyChannels` working properly in order to produce a dataset suitable for testing `NoisyChannels`. To address #83 I manually set the Z threshold to 3.29 (as suggested) and added a note so we don't forget.

Let me know what you think!

# Merge Checklist

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -42,10 +42,10 @@ def raw():
 def raw_clean(montage):
     """Return an `mne.io.Raw` object with no bad channels for use with tests.
 
-    This fixture downloads and reads in subject 1, run 1 from the Physionet
-    BCI2000 (eegbci) open dataset, interpolates its two bad channels (T10 & F8),
-    and performs average referencing on the data. Intended for use with tests
-    where channels are made artifically bad.
+    This fixture downloads and reads in subject 30, run 2 from the Physionet
+    BCI2000 (eegbci) open dataset, which contains no bad channels on an initial
+    pass of :class:`pyprep.NoisyChannels`. Intended for use with tests where
+    channels are made artifically bad.
 
     File attributes:
     - Channels: 64 EEG
@@ -57,17 +57,12 @@ def raw_clean(montage):
     """
     mne.set_log_level("WARNING")
 
-    # Download and read S001R01.edf from the BCI2000 dataset
-    edf_fpath = eegbci.load_data(1, 1, update_path=True)[0]
+    # Download and read S030R02.edf from the BCI2000 dataset
+    edf_fpath = eegbci.load_data(30, 2, update_path=True)[0]
     raw = mne.io.read_raw_edf(edf_fpath, preload=True)
     eegbci.standardize(raw)  # Fix non-standard channel names
 
-    # Interpolate the file's few bad channels to produce a clean dataset
+    # Set a montage for use with RANSAC
     raw.set_montage(montage)
-    raw.info["bads"] = ["T10", "F8"]
-    raw.interpolate_bads()
-
-    # Re-reference the data after interpolating bad channels
-    mne.set_eeg_reference(raw, 'average', copy=False, ch_type='eeg')
 
     return raw
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -14,15 +14,60 @@ def montage():
 
 @pytest.fixture(scope="session")
 def raw():
-    """Fixture for physionet EEG subject 4, dataset 1."""
+    """Return an `mne.io.Raw` object for use with unit tests.
+
+    This fixture downloads and reads in subject 4, run 1 from the Physionet
+    BCI2000 (eegbci) open dataset. This recording is quite noisy and is thus a
+    good candidate for testing the PREP pipeline.
+
+    File attributes:
+    - Channels: 64 EEG
+    - Sample rate: 160 Hz
+    - Duration: 61 seconds
+
+    This is only run once per session to save time downloading.
+
+    """
     mne.set_log_level("WARNING")
-    # load in subject 1, run 1 dataset
+
+    # Download and read S004R01.edf from the BCI2000 dataset
     edf_fpath = eegbci.load_data(4, 1, update_path=True)[0]
-
-    # using sample EEG data (https://physionet.org/content/eegmmidb/1.0.0/)
     raw = mne.io.read_raw_edf(edf_fpath, preload=True)
+    eegbci.standardize(raw)  # Fix non-standard channel names
 
-    # The eegbci data has non-standard channel names. We need to rename them:
-    eegbci.standardize(raw)
+    return raw
+
+
+@pytest.fixture(scope="session")
+def raw_clean(montage):
+    """Return an `mne.io.Raw` object with no bad channels for use with tests.
+
+    This fixture downloads and reads in subject 1, run 1 from the Physionet
+    BCI2000 (eegbci) open dataset, interpolates its two bad channels (T10 & F8),
+    and performs average referencing on the data. Intended for use with tests
+    where channels are made artifically bad.
+
+    File attributes:
+    - Channels: 64 EEG
+    - Sample rate: 160 Hz
+    - Duration: 61 seconds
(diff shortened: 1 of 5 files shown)


a-hurst:
Huh, looks like this had the unintended effect of making other sections of the test suite slower? Looking back at the old tests it looks like the current `test_findnoisychannels` makes direct modifications to the `raw` fixture that's used across tests, such that the `PrepPipeline` and `Reference` tests were taking a modified `raw` as input:

https://github.com/sappelhoff/pyprep/blob/bc46978b912737bc3b843438a3cbe1cd9d4dc33d/tests/test_find_noisy_channels.py#L15-L32

Since the above code tries to clean bad channels from the recording, I'm guessing that meant it took fewer iterations to re-reference during later tests. Regardless, I doubt that was intentional so I guess this PR fixes that as well!

tests/test_find_noisy_channels.py
```diff
@@ -1,155 +1,212 @@
...
+    # Make the signal for a different channel have a very low amplitude
+    low_dev_idx = (high_dev_idx - 1) if high_dev_idx > 0 else 1
+    raw_tmp._data[low_dev_idx, :] *= low_dev_factor
+
+    # Test detection of abnormally low-amplitude channels
+    # NOTE: The default z-score threshold (5.0) is too strict to allow detection
+    # of abnormally low-amplitude channels in some datasets. Using a relaxed Z
+    # threshold of 3.29 (p < 0.001, two-tailed) until a better solution is found.
+    nd = NoisyChannels(raw_tmp, do_detrend=False)
+    nd.find_bad_by_deviation(deviation_threshold=3.29)
```
sappelhoff, line 119: Could we document this in the MatPREP differences document? And should we wrap the original threshold of 5 in a `matlab_strict`  conditional? What do you think? I am all for improving Matprep where it makes sense (see also your issue https://github.com/VisLab/EEG-Clean-Tools/issues/26, and the median problem for which we should open an issue at some point, etc.) ... BUT for a 0.4 release I think it'd be great to get MatPrep equivalence, and not go beyond already (at least not without a matlab_strict option).

tests/conftest.py
```diff
@@ -14,15 +14,60 @@ def montage():
...
+    return raw
+
 
-    # using sample EEG data (https://physionet.org/content/eegmmidb/1.0.0/)
+@pytest.fixture(scope="session")
+def raw_clean(montage):
+    """Return an `mne.io.Raw` object with no bad channels for use with tests.
+
+    This fixture downloads and reads in subject 1, run 1 from the Physionet
+    BCI2000 (eegbci) open dataset, interpolates its two bad channels (T10 & F8),
```
sappelhoff, line 46: is there perhaps a file from that dataset that is clean to begin with? ... If you don't wanna test all of them, that's fine though. This is just a side note.

tests/test_find_noisy_channels.py
```diff
@@ -1,155 +1,212 @@
 """Test the find_noisy_channels module."""
 import numpy as np
+from numpy.random import RandomState
```
sappelhoff, line 3: what's the reason to do this instead of `np.random.RandomState` ?

tests/test_find_noisy_channels.py
```diff
@@ -1,155 +1,212 @@
...
+
+    """
+    raw_tmp = raw_clean_detrend.copy()
+    return raw_tmp
+
+
+def _generate_signal(fmin, fmax, timepoints, fcount=1):
+    """Generate an EEG signal from one or more sine waves in a frequency range."""
+    signal = np.zeros_like(timepoints)
+    for freq in rng.randint(fmin, fmax+1, fcount):
```
sappelhoff, line 50: if you want to use "rng" from outside of the function, instead of passing it into the function (which would probably be cleaner here and throughout), then I think you should make it `RNG` in capital letters, to make it slightly more obvious that it's somewhat like a "constant"

tests/conftest.py
```diff
@@ -14,15 +14,60 @@ def montage():
...
+
+    # Interpolate the file's few bad channels to produce a clean dataset
+    raw.set_montage(montage)
+    raw.info["bads"] = ["T10", "F8"]
+    raw.interpolate_bads()
 
-    # The eegbci data has non-standard channel names. We need to rename them:
-    eegbci.standardize(raw)
+    # Re-reference the data after interpolating bad channels
+    mne.set_eeg_reference(raw, 'average', copy=False, ch_type='eeg')
```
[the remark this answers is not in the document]
a-hurst: I honestly don't know, I haven't really used black formatting before! If you have a preference for double quotes over singles for this project I'll definitely stick to that.

tests/test_find_noisy_channels.py
a-hurst: Okay, will do.

tests/test_find_noisy_channels.py
a-hurst: I'm just modifying the test behaviour here rather than changing anything about NoisyChannels itself, I figured that a proper `matlab_strict` wrapper and explanation would be part of whatever PR ends up properly addressing #83 (which I'm fine leaving for after 0.4, I just wanted to make note of the issue as soon as I noticed it).

tests/conftest.py
a-hurst: Out of over 200 recordings, the script found 6 that were totally clean:

```
Clean recordings:
 - Subject 24 Run 2
 - Subject 30 Run 2
 - Subject 38 Run 1
 - Subject 58 Run 2
 - Subject 74 Run 2
 - Subject 91 Run 2
```

Glad I didn't try to do this manually!

tests/test_find_noisy_channels.py
sappelhoff: okay, makes sense :+1:

tests/conftest.py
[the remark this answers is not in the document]
a-hurst: From the look of it, PyPREP doesn't have any sort of automatic linting for Black enabled other than the 88 line length, is that something that should be fixed?

And that make sense! My general habit to use single-quotes for data (e.g., variable names, argument strings, key names, etc.) and double-quotes for strings that are meant to be read as text (e.g. docstrings, print messages, exception messages, etc.), but I'm happy to conform to whatever conventions a project uses. From what I can tell the Black style guide [prefers double-quotes to singles](https://black.readthedocs.io/en/stable/the_black_code_style.html#strings) (like you thought), so up to you.

tests/conftest.py
a-hurst: Fixed in the latest commit! For future reference, S024R02 had some sub-threshold high noise channels that broke the HF noise detection test. The median channel noisiness for S030R02 was like 30 times lower, so I used that instead.

tests/conftest.py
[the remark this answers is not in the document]
sappelhoff: argh, good point actually. This line is supposed to ensure in CI that all code is BLACK formatted: https://github.com/sappelhoff/pyprep/blob/bc46978b912737bc3b843438a3cbe1cd9d4dc33d/.github/workflows/python_tests.yml#L41

But it should say `black . --check` instead of `black .` --> the former does a check, whereas the latter just applies the formatting, which is then obviously lost because it's not committed.

And for devs, the development environment should be setup automatically such that BLACK formatting is applied prior to each commit (pre-commit hook, see https://github.com/sappelhoff/pyprep#installation)

I'll make a PR for that.

sappelhoff:
Thanks! I'll fix the formatting in https://github.com/sappelhoff/pyprep/pull/85 and make sure that the CI will check this in the future.

PR #85, opened 2021-05-11
CI: enforce black style

sappelhoff:
cc @a-hurst this PR's CI should fail now (only the first commit is expected to fail)

### DIFF
--- a/.github/workflows/python_tests.yml
+++ b/.github/workflows/python_tests.yml
@@ -38,7 +38,7 @@ jobs:
       run: |
         flake8 --docstring-convention numpy .
         check-manifest .
-        black .
+        black . --check
 
     - name: Test with pytest
       run: |
--- a/.github/workflows/python_tests.yml
+++ b/.github/workflows/python_tests.yml
@@ -39,6 +39,7 @@ jobs:
         flake8 --docstring-convention numpy .
         check-manifest .
         black . --check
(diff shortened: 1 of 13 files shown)


setup.cfg
```diff
@@ -69,6 +69,7 @@ exclude =
   examples/
 
 [isort]
+profile=black
```
sappelhoff, line 72: this is another important line to make isort (for auto formatting imports) work nicely with black (for auto formatting everything else)

sappelhoff:
you said you were using pylint, so this may be of interest to you: https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html?highlight=isort#pylint

.github/workflows/python_tests.yml
```diff
@@ -38,7 +38,8 @@ jobs:
       run: |
         flake8 --docstring-convention numpy .
         check-manifest .
-        black .
+        black . --check
+        isort . --check
```
sappelhoff, line 42: and this one doesn't hurt to add either

a-hurst (approved):
Looks good to me! Honestly, apart from my hefty use of single-quotes and the missing commas following final elements in lists, I'm amazed at how well we've followed strict Black style without being consciously aware of it.

I'll set up Black with pylint so I can stay consistent with this style going forwards.

PR #86, opened 2021-06-01
Fix large performance regression in _mat_quantile 

a-hurst:
# PR Description

I've been taking a break from PyPREP work for a bit to work on actual research projects, but yesterday I noticed while re-running PyPREP on my dataset with the latest commit that the time between RANSAC iterations seemed much longer than it used to. Breaking out cProfile and SnakeVis, I profiled the code for a single file and found that `_mat_quantile` and `_mat_iqr` were eating up a big chunk of the total runtime:

Screen Shot 2021-05-31 at 10 23 26 PM

Apparently, numpy's `apply_along_axis` function is famously slow, as are its `nanpercentile` and `nanquantile` functions. Luckily, I happened upon a [blog post](https://krstn.eu/np.nanpercentile()-there-has-to-be-a-faster-way/) of someone with a similar issue and adapted their solution to our use-case, so `_mat_iqr` and `_mat_quantile` are now fully vectorized while still handling NaNs properly!

On my local machine, this fix reduces the total time required to run the test suite from ~115 seconds down to about 50.

# Merge Checklist

### DIFF
--- a/pyprep/utils.py
+++ b/pyprep/utils.py
@@ -46,6 +46,8 @@ def _mat_round(x):
 def _mat_quantile(arr, q, axis=None):
     """Calculate the numeric value at quantile (`q`) for a given distribution.
 
+    Currently only supports 1-D or 2-D arrays.
+
     Parameters
     ----------
     arr : np.ndarray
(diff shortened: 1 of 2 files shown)


pyprep/utils.py
```diff
@@ -46,6 +46,8 @@ def _mat_round(x):
 def _mat_quantile(arr, q, axis=None):
     """Calculate the numeric value at quantile (`q`) for a given distribution.
 
+    Currently only supports 1-D or 2-D arrays.
```
sappelhoff, line 49: I would put this note under the `arr` parameter description. And also add a short check right at the beginning of the fun:

```python
if arr.ndim not in [1, 2]:
    raise ValueError(...)
```

pyprep/utils.py
```diff
@@ -73,28 +75,44 @@ def _mat_quantile(arr, q, axis=None):
...
-    q = np.asarray(q, dtype=np.float64)
-
-    if axis is not None and len(arr.shape) > 1:
-        # If an axis is specified, calculate quantiles along it
-        return np.apply_along_axis(_mat_quantile_1d, axis, arr, q)
+    # Sort the array in ascending order along the given axis (any NaNs go to the end)
+    # Return NaN if array is empty.
+    arr_sorted = np.sort(arr, axis=axis)
+    if arr_sorted.size == 0:
+        return np.NaN
```
sappelhoff, line 82: You could check this before sorting, right? no need to sort if you are going to return NaN anyhow.

pyprep/utils.py
```diff
@@ -73,28 +75,44 @@ def _mat_quantile(arr, q, axis=None):
...
-    # Make sure inputs are both Numpy arrays
-    arr = np.asarray(arr)
-    q = np.asarray(q, dtype=np.float64)
-
-    if axis is not None and len(arr.shape) > 1:
-        # If an axis is specified, calculate quantiles along it
-        return np.apply_along_axis(_mat_quantile_1d, axis, arr, q)
+    # Sort the array in ascending order along the given axis (any NaNs go to the end)
+    # Return NaN if array is empty.
+    arr_sorted = np.sort(arr, axis=axis)
```
sappelhoff, line 80: if axis=None, sort will "flatten" the array. You also flatten the array below again --> if you refactor this to do the sorting AFTER reshaping "...  data into a 2D array with the shape (num_axes, data_per_axis)", then you do one "flatten" less = tiny bit more speed

sappelhoff (commented):
cool, thanks :-) 

I didn't dive into the logic of the code ... we'll have to trust the blog post and your tests - but as those seem to pass, I think it's good to merge.

I left some minor comments.

pyprep/utils.py
a-hurst: For the array slicing at the bottom to work for both 1D and 2D inputs, 1D arrays need to be shaped into `(1, data_per_axis)` 2D arrays, so the second "flattening" still needs to happen as `sort` produces a 1D array with `axis=None`.

Also FYI I profiled the code again after the fix and ~95% of the time `_mat_quantile` takes now is due to the `sort` function (which is essential and I couldn't figure out any extra speed improvements for) so the total performance impact should be relatively small regardless

pyprep/utils.py
a-hurst: I was thinking that I wanted it to be coerced to `np.ndarray` format by `sort` already before checking size, but I realize now that `len(arr)` would work just as well for empty lists and arrays. Will fix in the next commit!

pyprep/utils.py
sappelhoff: Ah, makes sense - thanks! Also: if "sort" is the main culprit now, I think we are fine :-)

pyprep/utils.py
a-hurst: Fixed in the latest commit!

a-hurst:
> we'll have to trust the blog post and your tests - but as those seem to pass, I think it's good to merge.

This is where those MATLAB comparison tests really start to prove their worth! If we're still getting functionally identical arrays of values as MATLAB PREP, it's a lot easier to trust changes like this. They were definitely useful in debugging the new code.

sappelhoff:
Thank you @a-hurst - one more in :-)

PR #89, opened 2021-06-16
Use identical RNG state across RANSAC runs when matlab_strict enabled

### LINKED ISSUE #74: RANSAC random seeds work differently in PyPREP than in MATLAB PREP

a-hurst:
# PR Description

Closes #74 (at least as far as MATLAB equivalence is concerned). This PR modifies `NoisyChannels` to always use a copy of the random state object during RANSAC when `matlab_strict` is True. This results in the random state being identical for each run of `NoisyChannels.find_bad_by_ransac`, matching MATLAB PREP.

Also I realize I forgot to update the `whats_new.rst` for the last PR, so I did that here as well!

# Merge Checklist

### DIFF
--- a/docs/matlab_differences.rst
+++ b/docs/matlab_differences.rst
@@ -5,7 +5,7 @@
 Deliberate Differences from MATLAB PREP
 =======================================
 
-Although PyPREP aims to be a faithful reimplementaion of the original MATLAB
+Although PyPREP aims to be a faithful reimplementation of the original MATLAB
 version of PREP, there are a few places where PyPREP has deliberately chosen
 to use different defaults than the MATLAB PREP.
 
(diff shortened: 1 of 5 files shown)


sappelhoff (approved):
Thanks - also for adding that whatsnew entry!

sappelhoff (commented):
Just for clarification before merging: Does this enable us to do some matprep-artifact testing that we weren't able to cleanly do before?

a-hurst:
Nope: since the channel interpolation method still differs, the re-referencing process still ends up diverging numerically from MatPREP. This just makes it possible to add MatPREP comparisons for that.

sappelhoff:
okay :) then in it goes :rocket: thanks @a-hurst

PR #92, opened 2021-06-18
Ensure all reference channels are used during final average referencing

### LINKED ISSUE #90: PyPREP excludes interpolated NaN/flat/low-SNR channels from average referencing
While pulling together a fix for #77, I ended up catching another PyPREP/MatPREP difference I didn't notice before. Essentially, PyPREP re-uses a class attribute (`self.reference_channels`) to do the same job as two separate variables in MatPREP's `performReference` and `robustReference`, which ends up causing any channels flagged as bad-by-SNR, bad-by-NaN, or bad-by-flat on the initial NoisyChannels pass to be excluded from the average reference signal even after they've been interpolated.

Specifically, the `self.reference_channels` here:

https://github.com/sappelhoff/pyprep/blob/6abe0bcf1d1d40aefd2fe4cc41b200fbb0924ee4/pyprep/reference.py#L231-L238

gets reused here:

https://github.com/sappelhoff/pyprep/blob/6abe0bcf1d1d40aefd2fe4cc41b200fbb0924ee4/pyprep/reference.py#L114-L119

whereas in MATLAB PREP the former is a local variable and doesn't affect the latter.

a-hurst:
# PR Description

Closes #90. This PR fixes the issue by making `reference_channels` a local variable within `Reference.robust_reference()` so that it doesn't affect the `self.reference_channels` variable that's used to select reference channels during `Reference.perform_reference()`.

As an aside, is there any situation where it would make sense to only use a subset of all channels for final average reference calculation, post-interpolation? Wondering if the whole `reference_channels` parameter is a carry-over from MatPREP that only exists because EEGLAB doesn't have as nice a method of marking non-EEG channels as MNE. In MatPREP, channels are excluded from being sources for interpolation if they're not specified in the "reference channels" parameter, whereas PyPREP just uses all non-bad EEG channels, but I'm not sure if that would ever actually be a problem.

# Merge Checklist

### DIFF
--- a/docs/whats_new.rst
+++ b/docs/whats_new.rst
@@ -59,6 +59,7 @@ Bug
 - Fixed RANSAC to avoid making unnecessary signal predictions for known-bad channels, matching MATLAB behaviour and reducing RAM requirements, by `Austin Hurst`_ (:gh:`72`)
 - Fixed a bug in :meth:`NoisyChannels.find_bad_by_correlation` that prevented it from being able to handle channels with dropouts (intermittent flat regions), by `Austin Hurst`_ (:gh:`81`).
 - Fixed :class:`~pyprep.NoisyChannels` so that it always runs "bad channel by NaN" and "bad channel by flat" detection, preventing these channels from causing problems with other :class:`~pyprep.NoisyChannels` methods, by `Austin Hurst`_ (:gh:`79`)
+- Fixed :class:`~pyprep.Reference` so that channels are no longer excluded from final average reference calculation if they were originally bad by NaN, flat, or low SNR, by `Austin Hurst`_ (:gh:`92`)
 
 API
 ~~~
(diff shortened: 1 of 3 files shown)


a-hurst:
Also, check out the difference this makes in the MATLAB difference plot for the full PREP example in the docs. That example's not even running in MATLAB-strict mode, either!

Before this PR:
Screen Shot 2021-06-17 at 9 53 18 PM

After:
Screen Shot 2021-06-17 at 9 53 03 PM

sappelhoff (approved):
>  is there any situation where it would make sense to only use a subset of all channels for final average reference calculation, post-interpolation? 

I don't think so :thinking: you wouldn't even be able to call it an average reference if not all channels that are part of the data are averaged with respect to all of them.

This fix seems good, thanks for catching it!

a-hurst:
> I don't think so 🤔 you wouldn't even be able to call it an average reference if not all channels that are part of the data are averaged with respect to all of them.

That's what I thought. Wondering if that part of the default API (i.e., needing to specify "referenced" and "rereferenced" channels manually in the params dict instead of just defaulting to all EEG channels) might be worth changing before PyPREP officially leaves its "experimental" stage.

PR #93, opened 2021-06-24
Make noisy channel exclusion during Reference compatible with MATLAB PREP

### LINKED ISSUE #91: PyPREP interpolates more channels than MATLAB PREP during re-referencing
Another difference I caught trying to unit-test my code for #77: during re-referencing, after the initial pass that detects and removes bad-by-NaN/flat/low-SNR channels from the initial average reference, MATLAB PREP's rereferencing loop doesn't interpolate channels that are bad by SNR or dropout (including any bad-by-SNR channels that were flagged on the first pass).

Basically, in PyPREP we always update the list of total bad channels after each re-reference iteration using `NoisyChannels.get_bads()`, which returns all bad channel types. In MatPREP, it's updated by the following function:

```matlab
function ref = updateBadChannels(ref, noisy)
% Update the bad channel lists from ref based on bad channels in noisy
    ref.badChannelsFromNaNs = union(ref.badChannelsFromNaNs, ...
        noisy.badChannelsFromNaNs);
    ref.badChannelsFromNoData = union(ref.badChannelsFromNoData, ...
        noisy.badChannelsFromNoData);
    ref.badChannelsFromHFNoise = union(ref.badChannelsFromHFNoise, ...
        noisy.badChannelsFromHFNoise);
    ref.badChannelsFromCorrelation = union(ref.badChannelsFromCorrelation, ...
        noisy.badChannelsFromCorrelation);
    ref.badChannelsFromDeviation = union(ref.badChannelsFromDeviation, ...
        noisy.badChannelsFromDeviation);
    ref.badChannelsFromRansac = union(ref.badChannelsFromRansac, ...
        noisy.badChannelsFromRansac);
    ref.badChannelsFromDropOuts =  union(ref.badChannelsFromDropOuts, ...
        noisy.badChannelsFromDropOuts);
    ref.all = union(...
              union(ref.badChannelsFromNaNs, ref.badChannelsFromNoData), ...
              union( ...
                union(ref.badChannelsFromHFNoise, ref.badChannelsFromCorrelation), ...
              union( ...  
              union(ref.badChannelsFromDeviation,ref.badChannelsFromRansac), ...
                    ref.badChannelsFromRansac)));
```

As you can see,

1. Bad-by-SNR channels aren't updated with each iteration of re-referencing (make sense, since they're just the union of bad-by-HF and bad-by-correlation channels).
2. The set of all bads is the combination of all updated bad-by-NaN/flat/deviation/HF-noise/correlation/RANSAC channels instead of a union between `ref.all` and `noisy.all`. This means that any bad-by-SNR channels flagged during the first pass don't get carried over into the `ref.all` for future iterations if they aren't flagged bad by some other metric.
3. Dropout channels are updated with each iteration but aren't included in `ref.all`, possibly by mistake (see: the duplicate bad-by-RANSAC at the end).

Looking back at the [MatPREP commit where that function was introduced](https://github.com/VisLab/EEG-Clean-Tools/commit/21643cbecb4792148951e2286010121618af8023#diff-4c979a3b1247fbe34e47bbec4ca8eb5a8b60cf74eff2889108d8b8c17c7f7387), it looks like this function replaces simple updating of the full set of noisy channels so at least some of it seems to be deliberate?

Anyway, I'm not entirely sure how to best fix this because I can't easily tell whether PyPREP's logic or MatPREP's makes the most sense, and thus whether this should be fully or partly wrapped in `matlab_strict`. 3 certainly seems to be undesirable, since you probably wouldn't want a channel with intermittent flat regions being included in your estimated clean reference signal. What do you think?

a-hurst:
# PR Description

Closes #91. This PR does three main things:

- Changes `Reference.robust_reference` to stop permanently excluding channels that are flagged as bad-by-SNR prior to initial re-referencing.
- Changes `Reference.robust_reference` to exclude bad-by-dropout channels from the full set of bad channels when`matlab_strict` is `True`, matching MatPREP's (likely accidental) behaviour.
- Adds an `as_dict` argument to `NoisyChannels.get_bads` that returns bad channels by type as a dictionary in order to facilitate the above changes. In addition to being generally useful, it also saves a lot of dict construction in `reference.py`.

I still need to update the `whats_new.rst` and `matlab_differences.rst` here, but I figured given the time-zone difference that it would be better to get the code up for review before I went to bed. Thanks in advance for checking this over!

# Merge Checklist

### DIFF
--- a/docs/matlab_differences.rst
+++ b/docs/matlab_differences.rst
@@ -148,3 +148,43 @@ roughly mean-centered. and will thus produce similar values to normal Pearson
 correlation. However, to avoid making any assumptions about the signal for any
 given channel / window, PyPREP defaults to normal Pearson correlation unless
 strict MATLAB equivalence is requested.
+
+
+Differences in Robust Referencing
+---------------------------------
+
+During the robust referencing part of the pipeline, PREP tries to estimate a
+"clean" average reference signal for the dataset, excluding any channels
+flagged as noisy from contaminating the reference. The robust referencing
+process is performed using the following logic:
+
+1) First, an initial pass of noisy channel detection is performed to identify
+   channels bad by NaN values, flat signal, or low SNR: the data is then
+   average-referenced excluding these channels. These channels are subsequently
+   marked as "unusable" and are excluded from any future average referencing.
+
+2) Noisy channel detection is performed on a copy of the re-referenced signal,
+   and any newly detected bad channels are added to the full set of channels
+   to be excluded from the reference signal.
+
+3) After noisy channel detection, all bad channels detected so far are
+   interpolated, and a new estimate of the robust average reference is
+   calculated using the mean signal of all good channels and all interpolated
+   bad channels (except those flagged as "unusable" during the first step).
+
+4) A fresh copy of the re-referenced signal from Step 1 is re-referenced using
+   the new reference signal calculated in Step 3.
+
+5) Steps 2 through 4 are repeated until either two iterations have passed and
+   no new noisy channels have been detected since the previous iteration, or
+   the maximum number of reference iterations has been exceeded (default: 5).
+
+
+Exclusion of dropout channels
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+In MATLAB PREP, dropout channels (i.e., channels that have intermittent periods
+of flat signal) are detected on each iteration of the reference loop, but are
+currently not factored into the full set of "bad" channels to be interpolated.
+By contrast, PyPREP will detect and interpolate any bad-by-dropout channels
+detected during robust referencing.
--- a/docs/matlab_differences.rst
+++ b/docs/matlab_differences.rst
@@ -177,7 +177,7 @@ process is performed using the following logic:
 
 5) Steps 2 through 4 are repeated until either two iterations have passed and
    no new noisy channels have been detected since the previous iteration, or
-   the maximum number of reference iterations has been exceeded (default: 5).
+   the maximum number of reference iterations has been exceeded (default: 4).
 
 
 Exclusion of dropout channels
--- a/pyprep/reference.py
+++ b/pyprep/reference.py
@@ -241,10 +241,10 @@ class Reference:
             "bad_by_deviation": [],
             "bad_by_hf_noise": [],
             "bad_by_correlation": [],
-            "bad_by_SNR": noisy_detector.bad_by_SNR,
+            "bad_by_SNR": [],
             "bad_by_dropout": [],
             "bad_by_ransac": [],
-            "bad_all": self.unusable_channels,
+            "bad_all": _union(noisy_detector.bad_by_nan, noisy_detector.bad_by_flat),
         }
 
         # Get initial estimate of the reference by the specified method
--- a/pyprep/reference.py
+++ b/pyprep/reference.py
@@ -93,8 +93,14 @@ class Reference:
         self._extra_info = {}
         self.matlab_strict = matlab_strict
 
-    def perform_reference(self):
+    def perform_reference(self, max_iterations=4):
         """Estimate the true signal mean and interpolate bad channels.
+
+        Parameters
+        ----------
+        max_iterations : int, optional
+            The maximum number of iterations of noisy channel removal to perform
+            during robust referencing. Defaults to ``4``.
 
         This function implements the functionality of the `performReference` function
         as part of the PREP pipeline on mne raw object.
@@ -107,7 +113,7 @@ class Reference:
 
         """
         # Phase 1: Estimate the true signal mean with robust referencing
-        self.robust_reference()
+        self.robust_reference(max_iterations)
         # If we interpolate the raw here we would be interpolating
         # more than what we later actually account for (in interpolated channels).
         dummy = self.raw.copy()
@@ -165,11 +171,17 @@ class Reference:
 
         return self
 
-    def robust_reference(self):
+    def robust_reference(self, max_iterations=4):
         """Detect bad channels and estimate the robust reference signal.
 
         This function implements the functionality of the `robustReference` function
         as part of the PREP pipeline on mne raw object.
+
+        Parameters
+        ----------
+        max_iterations : int, optional
+            The maximum number of iterations of noisy channel removal to perform
+            during robust referencing. Defaults to ``4``.
 
         Returns
         -------
@@ -231,7 +243,6 @@ class Reference:
         raw_tmp = raw.copy()
         iterations = 0
         previous_bads = set()
-        max_iteration_num = 4
 
         while True:
             raw_tmp._data = signal_tmp * 1e-6
@@ -249,6 +260,7 @@ class Reference:
 
             # Specify bad channel types to ignore when updating noisy channels
             # NOTE: MATLAB PREP ignores dropout channels, possibly by mistake?
+            # see: https://github.com/VisLab/EEG-Clean-Tools/issues/28
             ignore = ["bad_by_SNR", "bad_all"]
             if self.matlab_strict:
                 ignore += ["bad_by_dropout"]
@@ -265,7 +277,7 @@ class Reference:
             if (
                 iterations > 1
                 and (len(bad_chans) == 0 or bad_chans == previous_bads)
-                or iterations > max_iteration_num
+                or iterations > max_iterations
             ):
                 break
             previous_bads = bad_chans.copy()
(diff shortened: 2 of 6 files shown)


pyprep/reference.py
```diff
@@ -134,17 +134,7 @@ def perform_reference(self):
...
-            "bad_by_flat": noisy_detector.bad_by_flat,
-            "bad_by_deviation": noisy_detector.bad_by_deviation,
-            "bad_by_hf_noise": noisy_detector.bad_by_hf_noise,
-            "bad_by_correlation": noisy_detector.bad_by_correlation,
-            "bad_by_SNR": noisy_detector.bad_by_SNR,
-            "bad_by_dropout": noisy_detector.bad_by_dropout,
-            "bad_by_ransac": noisy_detector.bad_by_ransac,
-            "bad_all": noisy_detector.get_bads(),
-        }
+        self.noisy_channels_before_interpolation = noisy_detector.get_bads(as_dict=True)
```
sappelhoff, line 137: this is nice :sunglasses:

pyprep/reference.py
```diff
@@ -260,7 +230,7 @@ def robust_reference(self):
         # Remove reference from signal, iteratively interpolating bad channels
         raw_tmp = raw.copy()
         iterations = 0
-        noisy_channels_old = []
+        previous_bads = set()
         max_iteration_num = 4
```
sappelhoff, line 234: perhaps adding a comment here that this number is derived from the paper? Otherwise it may appear magical

pyprep/reference.py
```diff
@@ -272,51 +242,43 @@ def robust_reference(self):
...
-                self.noisy_channels["bad_by_ransac"], noisy_detector.bad_by_ransac
-            )
-            self.noisy_channels["bad_all"] = _union(
-                self.noisy_channels["bad_all"], noisy_detector.get_bads()
-            )
-            logger.info("Bad channels: {}".format(self.noisy_channels))
+            noisy_new = noisy_detector.get_bads(as_dict=True)
+
+            # Specify bad channel types to ignore when updating noisy channels
+            # NOTE: MATLAB PREP ignores dropout channels, possibly by mistake?
```
sappelhoff, line 251: here it would be nice if you could open that issue on matprep and link to the issue from here.

If you don't want to do that, please just link to the specific comment where this is discussed in our own issue thread :)

sappelhoff (commented):
LGTM! Only minor comments for docs

pyprep/reference.py
a-hurst: It's actually a configurable parameter in MatPREP (4 is just the default): should we make it a parameter in PyPREP as well?

pyprep/reference.py
a-hurst: Just created an issue here: https://github.com/VisLab/EEG-Clean-Tools/issues/28, will add a reference in the code.

pyprep/reference.py
sappelhoff: Yes, I think that'd be best 👍

a-hurst:
@sappelhoff Okay, I've updated the docs and whats-new, and have addressed your comments. A couple questions before this gets merged:

- Any thoughts on whether there should be an API for configuring `max_iterations` at the `PrepPipeline` level? There's currently no API for configuring `NoisyChannels` thresholds from `PrepPipeline` either, so I wasn't sure what the expectation was for `Reference` settings.
- While writing out the `matlab_differences` explanation, I realized two weird things about the current logic we might want to change:
    - With the current changes, channels flagged as bad-by-SNR *before* initial average referencing are no no longer permanently flagged as "bad" after initial average referencing, the same as bad-by-deviation/noise/correlation/dropout/RANSAC channels. However, because they're included as part of the `self.unusable_channels` set, they're still always excluded from the calculated average reference signal! This means that initial bad-by-SNR channels are considered good enough to use when interpolating flagged bad channels, but not good enough to include under any circumstances in the average reference. This seems wrong, but is exactly what MATLAB PREP is doing. Any thoughts on whether we want to change this?
    - Currently, bad-by-dropout channels are flagged and removed if detected after initial average referencing (provided `matlab_strict` is False). However, if the signal isn't average-referenced to begin with, wouldn't the initial average referencing make any intermittent flat regions in the data no longer flat? As such, wouldn't we want bad-by-dropout channels to be flagged as unusable *prior* to initial referencing, the same as with bad-by-NaN and bad-by-flat channels?

sappelhoff:
> Any thoughts on whether there should be an API for configuring max_iterations at the PrepPipeline level? There's currently no API for configuring NoisyChannels thresholds from PrepPipeline either, so I wasn't sure what the expectation was for Reference settings

I think that if a parameter is configurable in MatPREP, and it's reasonably easy to make it configurable in PyPrep, then we SHOULD make it possible. These params should always have a reasonable default (like in MatPrep) though. On the other hand, let's not expose more params than MatPrep does (for now), unless we judge it very very important.

> This means that initial bad-by-SNR channels are considered good enough to use when interpolating flagged bad channels, but not good enough to include under any circumstances (even if interpolated) in the average reference

🤔 yeah, that sounds off. +1 for an issue about this 1. on pyprep and 2. on matprep. The one for pyprep can be marked to be resolved in version 0.5 --> version 0.4 being targeted more at feature completeness with matprep. If nobody ever reacts to this on matprep, we can do our own, non-matprep-strict thing in pyprep starting with 0.5.

> However, if the signal isn't average-referenced to begin with, wouldn't the initial average referencing make any intermittent flat regions in the data no longer flat? As such, wouldn't we want bad-by-dropout channels to be flagged as unusable prior to initial referencing, the same as with bad-by-NaN and bad-by-flat channels?

Yes, that's true, good point. What does matprep do here?

sappelhoff (commented):
there are three lines in this diff that are not being covered --> is it possible to reasonably do so with little changes/effort? If not, no problem.

docs/matlab_differences.rst
```diff
@@ -148,3 +148,43 @@ roughly mean-centered. and will thus produce similar values to normal Pearson
...
+   interpolated, and a new estimate of the robust average reference is
+   calculated using the mean signal of all good channels and all interpolated
+   bad channels (except those flagged as "unusable" during the first step).
+
+4) A fresh copy of the re-referenced signal from Step 1 is re-referenced using
+   the new reference signal calculated in Step 3.
+
+5) Steps 2 through 4 are repeated until either two iterations have passed and
+   no new noisy channels have been detected since the previous iteration, or
+   the maximum number of reference iterations has been exceeded (default: 5).
```
sappelhoff: isn't the default 4?

sappelhoff: ah sorry, I looked at an old commit. Ignore this please.

a-hurst:
> I think that if a parameter is configurable in MatPREP, and it's reasonably easy to make it configurable in PyPrep, then we SHOULD make it possible.

Okay, I'll add it to the `params` dict for now!

> +1 for an issue about this 1. on pyprep and 2. on matprep. The one for pyprep can be marked to be resolved in version 0.5 --> version 0.4 being targeted more at feature completeness with matprep. If nobody ever reacts to this on matprep, we can do our own, non-matprep-strict thing in pyprep starting with 0.5.

Sounds good, will do! Maybe this is a place for some experimentation, but my gut instinct would be to default to only marking initial bad-by-flat and bad-by-NaN channels as fully "unusable". I'd imagine that the original intent was to removing channels that contribute more noise than signal from contaminating the average reference, but given how sensitive the 'bad-by-correlation' default settings are (if over 1% of all correlation windows are below threshold, the whole channel is considered bad) and how that can change after average referencing, I don't think initial bad-by-SNR channels should be treated as permanently bad. Anyway, that can all be discussed in the new issue.

> Yes, that's true, good point. What does matprep do here?

It just ignores dropout channels completely during referencing: only bad-by-flat/NaN/SNR channels are permanently flagged and removed from the average reference at the onset, and due to the likely bug we've reported already they won't get identified or excluded on subsequent iterations either.

> there are three lines in this diff that are not being covered --> is it possible to reasonably do so with little changes/effort? If not, no problem.

The `matlab_strict` line will get covered in the EEGLAB interpolation PR when I can add the MatPREP comparison tests, but I'll try adding a test case to cover the other main ones!

a-hurst:
Okay, I've made codecov a bit happier now: I covered a line that only gets run if the input signal is completely clean, fixed a bug that prevented a ValueError from getting covered (it was throwing a ValueError earlier, so the test passed but the coverage failed), and worked around a known `coverage.py` bug where a `break` immediately after an `if` statement doesn't get counted as being run.

sappelhoff (approved):

sappelhoff:
Thanks @a-hurst

Issue #73, opened 2021-05-03
Extend PrepPipeline to allow performing PREP in stages


a-hurst:
While chugging along on the PyPREP/MatPREP comparison unit tests, I had a thought about the PyPREP API: since there are more-or-less three distinct stages to the PREP pipeline (adaptive line noise removal, robust re-referencing, interpolation of remaining bad channels), what if the `PrepPipeline` object had separate methods for each of these? The `fit()` method would still exist, but just run all of these in sequence.

Separate methods would be handy for situations like my own, where I basically copy/pasted the guts of `fit()` [into my own script](https://github.com/LBRF/DEMI_Analysis_Pipeline/blob/8534580133d260f7adc0259b4d85c4e2e4ea7d3f/_Preprocessing/eeg_pipeline.py#L259-L303) in order to do progress messages and plotting between CleanLine and robust referencing. Also for my script the project lead wanted the option to ignore bad channels instead of interpolating them, so I had to write extra code to revert interpolation and the names of bad channels instead of having the option to not interpolate in the first place.

Here's my basic proposal for `PrepPipeline` methods:
- `fit`: Performs the whole PREP pipeline, like ususal
- `remove_line_noise`: Performs adaptive line noise removal.
- `robust_reference`: Performs robust re-referencing without interpolation. Will default to printing a warning if `remove_line_noise` isn't done first.
- `interpolate_bads`: Interpolates any remaining bad channels and adjusts the average reference accordingly.

Let me know what you think!

yjmantilla:
I agree. Usually I dont do the remove line noise at all because the signals were already filtered.

sappelhoff:
I think I forgot how the robust re referencing works ... I am surprised that it can work without interpolation. Can you explain? 

Other than that, the api proposal sounds good to me. As long as we have good documentation an warnings in place to prevent users from accidentally screwing up their data, it should be fine.

yjmantilla:
The interpolation inside the robust referencing is always done, but it doesn't affect the real signal but a dummy-temporal one , that interpolation is only done to calculate the robust reference.

What I get is that at the end of robust referencing PREP once again searches for bad channels but this time he interpolates it on the final signal returned. This channels then become the "interpolated channels". 

Then prep once again searches for bad channels and what he finds is named "still_noisy_channels".

a-hurst:
> I understood that @a-hurst meant without the "final" interpolation.

[...]

PR #96, opened 2021-06-27
Implement EEGLAB-style interpolation in Reference for matlab_strict

### LINKED ISSUE #77: Interpolation of bad channels differs numerically between PyPREP and MatPREP
For whatever reason, MATLAB PREP uses a different method of interpolation for channel prediction during RANSAC than it does for interpolating channels during and after re-referencing.

For RANSAC, they use their own internal interpolation code (a copy of the private interpolation code from EEGLAB's `clean_rawdata` plugin), which happens to match MNE's `_make_interpolation_matrix` exactly. For all other interpolation, they use a wrapper around EEGLAB's public interpolation code, which seems to use a different method of spherical spline interpolation (different numeric results, no reference to Perrin '89 in the docstring).

Anyway, I can't think of any a priori justification for using different interpolation logic in one place over the other and my gut is to trust MNE's code more than EEGLAB's, but regardless it's something that needs to be fixed for `Reference` to produce numerically-identical results to MATLAB PREP for #58. According to the original manuscript it seems like they comparison-tested the methods and found they gave similar correlations with actual signals?

> The PREP pipeline uses the `spherical` option of EEGLAB `eeg_interp` function for channel interpolation. This function uses Legendre polynomials up through degree 7. To test this choice, we compared this interpolation method with the `v4` option and two other spherical interpolation functions, including the one used in RANSAC. **We applied a rank sum test on the correlation of the interpolated channel with the actual channel and found no significant difference between the three spherical interpolation methods**. The `v4` option performed significantly worse. The results were not sensitive to whether a block of channels was removed in the immediate vicinity of the bad channel or whether the interpolation was done in ICA space and projected back.

Sort of confused by "there were no significant differences, but `v4` was significantly worse", though. May be worth doing our own testing.

a-hurst:
# PR Description

Closes #77. This PR basically reimplements most of EEGLAB's `eeg_interp` method in Python so we can finally do MATLAB equivalence tests for the entirety of the `Reference` class (also added in this PR). This was... *not easy* to say the least, but it was necessary for the peace of mind that these MatPREP comparison CI tests allow (and also helped us identify a bunch of other small bugs/discrepancies along the way).

This is also the last numeric difference from MatPREP I'm aware of other than adaptive line noise removal (and differences between EEGLAB's montage-importing code and MNE's which result in differences in RANSAC, but that's not exactly a *PREP* divergence as much as an EEGLAB/MNE one).

Anyway, I've tried my best to make sense of and comment/document the re-implemented EEGLAB logic, but there are still a few areas I don't fully understand. If you have any insights, please let me know!

# Merge Checklist

### DIFF
--- a/docs/matlab_differences.rst
+++ b/docs/matlab_differences.rst
@@ -213,6 +213,10 @@ following robust referencing. However, for full numeric equivalence with
 MATLAB PREP, PyPREP will use a Python reimplementation of ``eeg_interp`` instead
 when the ``matlab_strict`` parameter is set to ``True``.
 
+
+References
+----------
+
 .. [1] Perrin, F., Pernier, J., Bertrand, O. and Echallier, JF. (1989).
    Spherical splines for scalp potential and current density mapping.
    Electroencephalography Clinical Neurophysiology, Feb; 72(2):184-7.
--- a/docs/matlab_differences.rst
+++ b/docs/matlab_differences.rst
@@ -188,3 +188,31 @@ of flat signal) are detected on each iteration of the reference loop, but are
 currently not factored into the full set of "bad" channels to be interpolated.
 By contrast, PyPREP will detect and interpolate any bad-by-dropout channels
 detected during robust referencing.
+
+
+Bad channel interpolation
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+MATLAB PREP uses EEGLAB's internal ``eeg_interp`` method of spherical spline
+interpolation for interpolating identified bad channels during robust reference
+estimation and (if enabled) immediately after the robust reference signal is
+applied in order to remove any remaining detected bad channels once referencing
+is complete.
+
+However, ``eeg_interp``'s method of spherical interpolations differs quite a bit
+numerically from MNE's implementation as well as the interpolation method used
+by MATLAB PREP for RANSAC predictions, both of which are numerically identical
+and based directly on the formulas in Perrin et al. (1989) _[1]. ``eeg_interp``
+seems to use a modified variation of the Perrin et al. method, but diverges in
+a number of ways that are not clearly documented or cited in the code.
+
+To keep with the more established method of spherical interpolation and stay
+consistent with the interpolation code used in RANSAC, PyPREP defaults to using
+MNE's :meth:`~mne.io.Raw.interpolate_bads` method for interpolation during and
+following robust referencing. However, for full numeric equivalence with
+MATLAB PREP, PyPREP will use a Python reimplementation of ``eeg_interp`` instead
+when ``matlab_strict`` is set to ``True``.
+
+.. [1] Perrin, F., Pernier, J., Bertrand, O. and Echallier, JF. (1989).
+   Spherical splines for scalp potential and current density mapping.
+   Electroencephalography Clinical Neurophysiology, Feb; 72(2):184-7.
--- a/docs/matlab_differences.rst
+++ b/docs/matlab_differences.rst
@@ -9,10 +9,10 @@ Although PyPREP aims to be a faithful reimplementation of the original MATLAB
 version of PREP, there are a few places where PyPREP has deliberately chosen
 to use different defaults than the MATLAB PREP.
 
-To override these differerences, you can set the ``matlab_strict`` argument to
-:class:`~pyprep.PrepPipeline`, :class:`~pyprep.Reference`, or
-:class:`~pyprep.NoisyChannels` as ``True`` to match the original PREP's
-internal math.
+To override these differerences, you can set the ``matlab_strict`` parameter
+for :class:`~pyprep.PrepPipeline`, :class:`~pyprep.Reference`, or
+:class:`~pyprep.NoisyChannels` to ``True`` in order to match the original
+PREP's internal math.
 
 .. contents:: Table of Contents
     :depth: 3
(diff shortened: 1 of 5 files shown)


a-hurst:
Note: if the docstring for `_eeglab_calc_g` seems a little opaque, that's because it's more-or-less a direct copy of the docstring for MNE's [equivalent internal function](https://github.com/mne-tools/mne-python/blob/3078533f8704820a168797578f5b47ddff4be347/mne/channels/interpolation.py#L35-L52). I'm not sure I understand the logic behind the actual math well enough to write anything more informative.

docs/matlab_differences.rst
```diff
@@ -188,3 +188,31 @@ of flat signal) are detected on each iteration of the reference loop, but are
...
+and based directly on the formulas in Perrin et al. (1989) _[1]. ``eeg_interp``
+seems to use a modified variation of the Perrin et al. method, but diverges in
+a number of ways that are not clearly documented or cited in the code.
+
+To keep with the more established method of spherical interpolation and stay
+consistent with the interpolation code used in RANSAC, PyPREP defaults to using
+MNE's :meth:`~mne.io.Raw.interpolate_bads` method for interpolation during and
+following robust referencing. However, for full numeric equivalence with
+MATLAB PREP, PyPREP will use a Python reimplementation of ``eeg_interp`` instead
+when ``matlab_strict`` is set to ``True``.
```
sappelhoff, line 214:
```suggestion
when the ``matlab_strict`` parameter is set to ``True``.
```

tests/test_matprep_compare.py
```diff
@@ -291,3 +349,48 @@ def test_all_bads(self, pyprep_noisy, matprep_noisy):
...
+    """Compare the results of Reference to the equivalent MatPREP code.
+
+    These comparisons use input data that's already had adaptive line noise
+    removal done to the signal, so any differences in results will be due to
+    differences in the robust referencing code itself.
+
+    """
+
+    # TODO: once final interpolation is separated out in PyPREP, add test just
+    # for interpolation code
```
sappelhoff, line 364: Can you open an issue for that once this PR is merged? That'd make it easier to track, as code comments are easily forgotten

sappelhoff (commented):
Thanks! Nice to see all the equivalence tests :-)

One more general comments about the EEGLAB "translation" to Python --> i'd appreciate if you could go to the EEGLAB GitHub repo and create "permanent links" to the parts of the codebase that you used for the translation.

For example, you have these two lines of code:

```python
  # Subtract all the Euclidian electrode distances from 1 (why?)
    EI = np.ones([n_to, n_from]) - elec_dists
```

This would be significantly enhanced (from a dev/docs perspective) if you add one more comment line like this (note, that's just an example):

```python
# https://github.com/sccn/eeglab/blob/167dfc8766a253619fc586848c704665e39bce9f/Contents.m#L8
```

You can also shorten the hash to 7 characters, instead of using the full hash:

`https://github.com/sccn/eeglab/blob/167dfc8/Contents.m#L8`

a-hurst:
> One more general comments about the EEGLAB "translation" to Python --> i'd appreciate if you could go to the EEGLAB GitHub repo and create "permanent links" to the parts of the codebase that you used for the translation.

Do you mean per function, or per individual chunk of code (e.g. the `EI = np.ones([n_to, n_from]) - elec_dists` line you mentioned)? For the latter, it's a little messy since I refactored the code a bit for readability & consistency with MNE's versions of these functions, but I can certainly try!

tests/test_matprep_compare.py
a-hurst: I've updated the first comment in #73 to mention this, does that work? https://github.com/sappelhoff/pyprep/issues/73

sappelhoff: yes thank you - that's good enough!

sappelhoff:
> Do you mean per function, or per individual chunk of code (e.g. the EI = np.ones([n_to, n_from]) - elec_dists line you mentioned)? For the latter, it's a little messy since I refactored the code a bit for readability & consistency with MNE's versions of these functions, but I can certainly try!

I don't have a system in mind, so I'd leave this up to you and recommend to do it in a "reasonable way". For example if you take something from one file in EEGLAB, then put a link ... if the file you draw from changes further down your code, put in a new link.

The readers can then be expected to follow your link and pick up the pieces themselves. that is: not every line of code must be mapped :+1:

docs/matlab_differences.rst
```diff
@@ -188,3 +188,31 @@ of flat signal) are detected on each iteration of the reference loop, but are
...
+a number of ways that are not clearly documented or cited in the code.
+
+To keep with the more established method of spherical interpolation and stay
+consistent with the interpolation code used in RANSAC, PyPREP defaults to using
+MNE's :meth:`~mne.io.Raw.interpolate_bads` method for interpolation during and
+following robust referencing. However, for full numeric equivalence with
+MATLAB PREP, PyPREP will use a Python reimplementation of ``eeg_interp`` instead
+when the ``matlab_strict`` parameter is set to ``True``.
+
+.. [1] Perrin, F., Pernier, J., Bertrand, O. and Echallier, JF. (1989).
```
sappelhoff, line 216: you may want to put a "references" heading above this one. But this is a style preference, so only do it if you agree that it's nicer.

https://sphinxcontrib-bibtex.readthedocs.io/en/latest/index.html would probably be overkill for the few citations we have, but that'd be the canonical way to solve this.

a-hurst:
> I don't have a system in mind, so I'd leave this up to you and recommend to do it in a "reasonable way". For example if you take something from one file in EEGLAB, then put a link ... if the file you draw from changes further down your code, put in a new link.

Okay, I put line references to the definitions of the functions I reimplemented. Hopefully that's enough for anyone curious in the future!

docs/matlab_differences.rst
a-hurst: Added it in. I think the docs look a bit better with it in there!

sappelhoff (approved):
thanks! :)

sappelhoff:
@a-hurst this question on MNE discourse sounds like you could answer it: https://mne.discourse.group/t/montage-setting-differences-in-mne-vs-eeglab/3695/2

Issue #15, opened 2020-04-04
full PREP example should be split


sappelhoff:
Currently the full PREP example does both, showing *how* it can be used, and *validating* the results against the Matlab results.

https://pyprep.readthedocs.io/en/latest/auto_examples/index.html

I think we should rather have two examples:

- one example where PREP usage is shown more extensively and with more explanations than currently
- and another example with a focus on validation against the Matlab results (here, the plotting code should be improved, because currently it's very messy to read and probably lots of unneeded code duplication)

sappelhoff:
Note that regarding the validation part of the example, we should also rely on @a-hurst's https://github.com/a-hurst/matprep_artifacts/tree/master

instead of the old saved data. As for some reason, things have become out of sync and it looks terrible now: https://pyprep.readthedocs.io/en/latest/auto_examples/run_full_prep.html#sphx-glr-auto-examples-run-full-prep-py

PR #99, opened 2021-07-03
[WIP] Split PrepPipeline into separate methods, make final interpolation optional

### LINKED ISSUE #73: Extend PrepPipeline to allow performing PREP in stages

a-hurst:
# PR Description

(Eventually) closes #73. Currently, this PR:

- Adds separate `remove_line_noise` and `robust_reference` methods to `PrepPipeline`.
- Makes final bad channel interpolation optional.
- Removes a bunch of unused matplotlib-based MATLAB comparison code in the `PrepPipeline` unit tests.
- Removes support for notch filter types other than `spectrum_fit`, since you can now just filter line noise however you want and then run `PrepPipeline.robust_reference` if you want to use another kind. This vastly simplifies the demands on the `filter_kwargs` API and should make the `PrepSettings` filtering options much easier to document.

There's still definitely more cleanup to do, but I figured it'd be good to get the core of this up for review sooner than later!

# Merge Checklist


a-hurst:
Whoops, looks like one of the examples relies on an undocumented attribute I removed for the sake of RAM (`self.EEG_new`). Will address that tomorrow.

sappelhoff (commented):
thanks, it's great that we don't need the matplotlib stuff in the tests anymore. Also reminds me of #15

sappelhoff:
It is possible that I introduced these conflicts :confounded: it's because I worked on dbe2062d3c7d50a0d87775e1111062ef854c366f before pulling `master` ... the pulled master, rebased, and force pushed. That was a bad idea - I hope nothing serious got destroyed.

a-hurst:
> It is possible that I introduced these conflicts 😖 it's because I worked on [dbe2062](https://github.com/sappelhoff/pyprep/commit/dbe2062d3c7d50a0d87775e1111062ef854c366f) before pulling `master` ... the pulled master, rebased, and force pushed. That was a bad idea - I hope nothing serious got destroyed.

No worries, I'll rebase before I push any more work! I've set this aside for a few days while tackling some other projects, but I'll try and get this wrapped up within the next week or so :)

a-hurst:
Okay @sappelhoff, I've gotten the `PrepPipeline` API rewrite to a point I think I'm happy with (though I'd love to hear your thoughts and @yjmantilla's). Some of the attributes aren't documented yet and much of it is untested, but I figured it'd be best to handle that side of things once I was sure we were all happy with the changes.

My main changes here are:

- Renamed `EEG_before_interpolation` to `EEG_post_reference` and `reference_before_interpolation` to `robust_reference_signal`
- Added a getter method `current_reference_signal` that returns `reference_before_interpolation` if interpolation hasn't been done, or the post-interpolation average reference (previously `reference_after_interpolation`) if it has (and `None` if neither have been performed, though that could be changed).
- Moved all the loose `noisy_channels_original`, `noisy_channels_before_interpolation`, `bad_before_interpolation`, `still_noisy_channels`, etc. attributes into two dicts: `noisy_info` (which contains the full noisy channel dicts) and `bad_channels` (which just contains the bad channel names), each with the keys 'original', 'post-reference', and 'post-interpolation'. This makes things easier to document and has the advantage of a more consistent/informative naming scheme.
- Added `current_noisy_info` and `remaining_bad_channels` attributes that get the current noisy info and bad channel names for the pipeline (will retrieve post-interpolation values if interpolation was used, otherwise returns the pre-interpolation values). These allow for people to try enabling/disabling interpolation for their data without having to rewrite their code to access different attributes.
- Added a new `get_raw` method, which is a more flexible version of the current `raw` getter that allows retrieving the full `mne.io.Raw` object with the EEG data from any given stage in the pipeline.

Let me know what you think!

examples/run_full_prep.py
```diff
@@ -168,7 +172,7 @@
 
 EEG_new_matlab = sio.loadmat(fname_mat2)
 EEG_new_matlab = EEG_new_matlab["save_data"]
-EEG_new = prep.EEG_new
+EEG_new = removeTrend(prep.EEG_raw, sample_rate=prep.sfreq) * 1e6
```
sappelhoff, line 175: @a-hurst could it be that you didn't rebase this PR or did a rebase that went wrong? Because we just remove this in #102

sappelhoff:
I didn't take a look at the diff because I am not sure what diff will stay after https://github.com/sappelhoff/pyprep/pull/99#discussion_r683235569 is resolved, but I still have concerns over these two points (I am commenting only based on your summary comment):

> returns reference_before_interpolation if interpolation hasn't been done, or the post-interpolation average reference (previously reference_after_interpolation) if it has

and

> (will retrieve post-interpolation values if interpolation was used, otherwise returns the pre-interpolation values

Could we not just return a dict in both cases, with the keys `pre_interpolation` and `post_interpolation` ... and the values for these keys being `None` or empty dicts if these values have not yet been computed? I think that'd be clearer. An alternative would be to send a log message upon the getter call that says something like: "you are getting the post_interpolation values". The minimum would be to have an attribute that can clearly tell you whether the prep object is "pre" or "post" interpolation, but we might already have something like that, not sure right now.

What do you think?

examples/run_full_prep.py
a-hurst: Nope, I somehow missed the full prep example during #102 so I'm fixing it here. My bad!

a-hurst:
> Could we not just return a dict in both cases, with the keys `pre_interpolation` and `post_interpolation` ... and the values for these keys being `None` or empty dicts if these values have not yet been computed? I think that'd be clearer. An alternative would be to send a log message upon the getter call that says something like: "you are getting the post_interpolation values". The minimum would be to have an attribute that can clearly tell you whether the prep object is "pre" or "post" interpolation, but we might already have something like that, not sure right now.

That could work, though I still think it'd be useful to have an attribute like `current_reference_signal` so there's a consistent way to get the final reference signal regardless if interpolation was enabled or disabled. For the last project we did with PyPREP we tried the analysis pipeline both with and without final bad channel interpolation, so I think that having an API where you can toggle interpolation on and off without needing to change any attributes you access afterwards would be handy.

Instead of attributes, maybe methods in the style of the new `get_raw` function would be clearer? For example, a `get_noisy_info` method that returns the most recent noisy info available by default (like `current_noisy_info` does now), but also has an argument that lets you specify explicitly whether you want the original, post-reference, or post-interpolation noisy info.

Keep in mind that for the `PrepPipeline` API I decided against exposing interpolation as a separate method and instead made it an optional flag to `robust_reference()`, so it should hopefully be clear to users based on the settings they chose (and the documentation) whether the "current reference signal" reflects interpolated data or not.

examples/run_full_prep.py
a-hurst: To clarify: beforehand, these `PrepPipeline` EEG data attributes were all in microVolts which matches EEGLAB convention, so the `* 1e6` wasn't needed to compare with MATLAB results. Now that those attributes are in MNE-standard Volts, we need to convert them to uV to compare with EEGLAB data the way we do here.

sappelhoff: Thanks!

pyprep/prep_pipeline.py
```diff
@@ -165,75 +163,193 @@ def __init__(
...
+
+        Valid pipeline stages include 'unprocessed' (the raw data prior to running
+        the pipeline), 'filtered' (the data following adaptive line noise
+        removal), 'post-reference' (the data after robust referencing, prior to any
+        bad channel interpolation), and 'post-interpolation' (the data after robust
+        referencing and bad channel interpolation).
+
+        Parameters
+        ----------
+        stage : str, optional
```
sappelhoff, line 220: I think I'd do something like this:

```suggestion
        stage : {"unprocessed", "filtered", "post-reference", "post-interpolation"}, optional
```

pyprep/prep_pipeline.py
```diff
@@ -165,75 +163,193 @@ def __init__(
...
-                    Fs=self.sfreq,
-                    freqs=linenoise,
-                    **self.filter_kwargs,
+        eeg_data = self.raw_eeg._data  # Default to most recent stage of pipeline
+        if stage:
+            eeg_data = stages[stage.lower()]
+            if not eeg_data:
+                raise ValueError(
+                    "Could not retrieve {stage} data, as that stage of the pipeline "
+                    "has not yet been performed."
```
sappelhoff, line 251:
```suggestion
                    f"Could not retrieve {stage} data, as that stage of the pipeline "
                    "has not yet been performed."
```

sappelhoff (commented):
> Instead of attributes, maybe methods in the style of the new get_raw function would be clearer? For example, a get_noisy_info method that returns the most recent noisy info available by default (like current_noisy_info does now), but also has an argument that lets you specify explicitly whether you want the original, post-reference, or post-interpolation noisy info.

yes! That sounds good to me, I like the new `get_raw` :+1: But I think the default should be `"current"`, not `"None"`, to be more explicit.

I approve of the remaining changes, but I am slowly losing my grasp on the potential workflows we can have with our different classes and their methods :innocent: it's been too long that I actually worked with this code, I think.

a-hurst:
> yes! That sounds good to me, I like the new `get_raw` 👍 But I think the default should be `"current"`, not `"None"`, to be more explicit.

Great! I'll get to work on this, as well as proper tests and docs for everything new. In that case, for the sake of simplicity I'm going to leave the internal noisy info and reference signal attributes undocumented so that the new `get_x` methods are the one clear official way for users to get at that data.

> I approve of the remaining changes, but I am slowly losing my grasp on the potential workflows we can have with our different classes and their methods 😇 it's been too long that I actually worked with this code, I think.

Hopefully cleaning up the example scripts for 0.4 will refresh us all on that front!

pyprep/prep_pipeline.py
a-hurst: That causes the line to go beyond 88 characters, is line wrap for argument types something that's supported by Numpy docstyle?

pyprep/prep_pipeline.py
a-hurst: Whoops, nice catch!

PR #110, opened 2022-01-16
Improve documentation for PyPREP/MatPREP highpass differences

### LINKED ISSUE #107: filt vs filtfilt (MATLAB implementation)
@sappelhoff So I'm currently taking a grad course on signal processing, and I think I finally understand the key difference between MNE's highpass filter and EEGLAB's: MNE defaults to applying the filter using a zero-phase method that doesn't result in any phase shift in the signal (the equivalent of `filtfilt` in MATLAB/Scipy), whereas the EEGLAB filter function I re-implemented here only does a single forward pass (using `lfilter`), so you end up with phase shift in the signal.

As far as I can tell, other than theoretically being faster (which it isn't in the current implementation) there's absolutely no reason to use a forwards-only filter like EEGLAB does instead of the backwards-forwards zero-phase approach used by MNE. I'm guessing this is a weird legacy carry-over for EEGLAB from the days when people were running EEG analyses on Pentium 3's and PowerPC G4's so CPU cycles were at a major premium. I'll update the "differences" docs to reflect this when I get a chance!

_Originally posted by @a-hurst in https://github.com/sappelhoff/pyprep/issues/71#issuecomment-973060655_

a-hurst:
# PR Description

Closes #107. This PR updates the docs to contain a better explanation of the differences between PyPREP's and MatPREP's trend removal filters (and why the MNE method is technically preferable).

Wasn't sure if this warranted an addition in the `whats_new.rst`, but if it does I can definitely add a line!

# Merge Checklist

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -4,6 +4,7 @@ flake8
 flake8-docstrings
 isort
 matplotlib
+mne[data]
 numpydoc
 pillow
 pre-commit
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -4,7 +4,8 @@ flake8
 flake8-docstrings
 isort
 matplotlib
-mne[data]
+tdqm
+pooch
 numpydoc
 pillow
 pre-commit
--- a/docs/matlab_differences.rst
+++ b/docs/matlab_differences.rst
@@ -33,9 +33,9 @@ In MATLAB PREP, the default method of trend removal is to use EEGLAB's
 data. MNE's :func:`mne.filter.filter_data` offers similar functionality, but
 with two key differences:
 
-1) MNE's method of FIR filter design is slightly different, resulting in
-   slightly higher FIR filter orders in EEGLAB than in MNE for the same input
-   values.
+1) MNE's method of FIR filter design has minor differences, resulting in
+   slightly lower FIR filter orders than EEGLAB for the same input
+   values (e.g. 845 instead of EEGLAB's 847).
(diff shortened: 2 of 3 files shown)


sappelhoff (approved):
perfect :-) Thanks a lot!

a-hurst:
Er, looks like MNE needs an additional module to download test datasets now? Should I add that to this PR as well?

sappelhoff:
ah right :thinking: yes please - we can just add it here: https://github.com/sappelhoff/pyprep/blob/master/requirements-dev.txt

sappelhoff:
although `pooch` may soon be installed automatically when we install `mne`, see: https://github.com/mne-tools/mne-python/pull/10199

but no harm in adding it now to merge this PR, and then removing it again some time in the future when we don't need to declare it explicitly anymore.

a-hurst:
Whoops, looks like matplotlib's changed its API on us since last update too. I'll try to tackle that as well.

requirements-dev.txt
```diff
@@ -4,6 +4,7 @@ flake8
 flake8-docstrings
 isort
 matplotlib
+mne[data]
```
sappelhoff, line 7: I prefer to be explicit here:

```suggestion
tqdm
pooch
```

a-hurst: Ah, okay. I figured `mne[data]` would future-proof against future MNE changes to data downloading, but being explicit manse sense.

sappelhoff:
Thank you :)

PR #112, opened 2022-03-21
MRG: Add mne logger

### LINKED ISSUE #105: Use the MNE logger to set the verbosity
I have been using PyPREP a bit (version 0.3.1, pip install) and I really like it for now.
Would it be possible to add a verbose argument to limit the prints? Alternatively, moving the print to a logger with a `set_log_level` function would work out too.

Example: `print("Executing RANSAC\nThis may take a while, so be patient...")`

When running a preprocessing pipeline with several steps, one of which is PyPREP, the relevant logging from the pipeline can be completely overwhelmed by the prints from PyPREP, especially if you run it as I do on e.g. 40 cores at once.

mscheltienne:
Closes #105.

I had this on my namespace for a while... (Oct 22. 2021) and I didn't take the time to finish it. 
If I recall correctly, it's missing the ProgressBar class for the 10%... 20%... 30%... print output by the RANSAC (probably the most annoying one when I run this on 40 cores 😅). 
I'll double-check the diff, make sure I didn't miss anything and add the ProgressBar for RANSAC.

Mathieu

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -40,6 +40,10 @@
       {
         "affiliation": "Institute of Electric Power Systems, Otto von Guericke University Magdeburg, Magdeburg, Germany",
         "name": "Jonte Dancker"
+      },
+      {
+        "affiliation": "Fondation Campus Biotech Geneva, Geneva, Switzerland",
+        "name": "Mathieu Scheltienne"
       }
     ],
     "keywords": [
--- a/docs/whats_new.rst
+++ b/docs/whats_new.rst
@@ -12,6 +12,7 @@ People who contributed to this software across releases (in **alphabetical order
 * `Adam Li`_
 * `Yorguin Mantilla`_
 * `Christian O'Reilly`_
+* `Mathieu Scheltienne`_
 * `Victor Xiang`_
 
 .. _whats_new:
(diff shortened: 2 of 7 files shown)


mscheltienne:
Also have to fix a couple of logs, as it seems I just did a `replace`

mscheltienne:
image

Looks like this is working. And it does get disabled if I change the loglevel of MNE.

sappelhoff (commented):
LGTM, thanks!

Please:

- add yourself here: https://github.com/sappelhoff/pyprep/blob/4a91d31782717c60e7c9e8814f68a83645f54475/.zenodo.json#L43
- and to the list of authors here: https://github.com/sappelhoff/pyprep/blob/master/docs/whats_new.rst (don't forget to also put a link at the bottom of the doc)
- finally, please prepare a short changelog entry (for syntax, see the previous entries)

We can ignore the codecov check.
Previously there was some RNG issue in the tests, that has now disappeared ... still a bit weird - but in any case unrelated to your changes.

mscheltienne:
Done

sappelhoff (approved):

sappelhoff:
Thanks @mscheltienne

PR #148, opened 2024-06-17
Fixes for Numpy 2

### LINKED ISSUE #147: CI: Fixes needed for Numpy 2.0
Apparently we need to introduce some fixes for Numpy 2.0

https://github.com/sappelhoff/pyprep/actions/runs/9541812770/job/26295569011#step:9:857

sappelhoff:
- closes #147

### DIFF
--- a/pyprep/find_noisy_channels.py
+++ b/pyprep/find_noisy_channels.py
@@ -1,6 +1,4 @@
 """finds bad channels."""
-from copy import copy
-
 import mne
 import numpy as np
 from mne.utils import check_random_state, logger
@@ -579,7 +577,14 @@ class NoisyChannels:
         exclude_from_ransac = (
             self.bad_by_correlation + self.bad_by_deviation + self.bad_by_dropout
         )
-        rng = copy(self.random_state) if self.matlab_strict else self.random_state
+
+        if self.matlab_strict:
+            random_state = self.random_state.get_state()
+            rng = np.random.RandomState()
+            rng.set_state(random_state)
+        else:
+            rng = self.random_state
+
         self.bad_by_ransac, ch_correlations_usable = find_bad_by_ransac(
             self.EEGFiltered,
             self.sample_rate,
--- a/pyprep/removeTrend.py
+++ b/pyprep/removeTrend.py
@@ -170,5 +170,5 @@ def runline(y, n, dn):
         (np.multiply(np.arange(n + 1, n + npts + 1), a) + b), (npts, 1)
     )
     for i in range(0, len(y_line)):
-        y[i] = y[i] - y_line[i]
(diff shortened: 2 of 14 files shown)


sappelhoff:
One of the issues is here: On MacOS it seems that the RNG *does* change the seed for some reason:

https://github.com/sappelhoff/pyprep/blob/e4840d9c31ccd880e427448edd8317ff9f6f7909/tests/test_find_noisy_channels.py#L249-L254

cc @a-hurst

sappelhoff:
with the most recent push, it turns out that the CI fails on all operating systems, so at least it's not something MacOS specific, after all.

pyprep/find_noisy_channels.py
```diff
@@ -579,7 +577,14 @@ def find_bad_by_ransac(
...
         )
-        rng = copy(self.random_state) if self.matlab_strict else self.random_state
+
+        if self.matlab_strict:
+            random_state = self.random_state.get_state()
+            rng = np.random.RandomState()
+            rng.set_state(random_state)
+        else:
+            rng = self.random_state
+
```
sappelhoff, line 587: This is one of the key fixes, for some reason.

pyprep/removeTrend.py
```diff
@@ -170,5 +170,5 @@ def runline(y, n, dn):
         (np.multiply(np.arange(n + 1, n + npts + 1), a) + b), (npts, 1)
     )
     for i in range(0, len(y_line)):
-        y[i] = y[i] - y_line[i]
+        y[i] = y[i] - y_line[i, 0]
```
sappelhoff, line 173: this is suppressing a future warning, because `y_line` is of shape (N, 1)

pyprep/utils.py
```diff
@@ -232,7 +232,7 @@ def _eeglab_fir_filter(data, filt):
         )
 
     # Finish filtering data, using padding at end to calculate final values
-    end_pad_idx = np.zeros(pad_len, dtype=np.uint8) + (n_samples - 1)
+    end_pad_idx = np.zeros(pad_len, dtype=np.uint) + (n_samples - 1)
```
sappelhoff, line 235: luckily, changing from uint8 to uint does not change anything ... probably because this was done already in the background before numpy 2.0

tests/conftest.py
```diff
@@ -77,7 +77,7 @@ def make_random_mne_object(
     n_freq_comps=5,
     freq_range=[10, 60],
     scale=1e-6,
-    RNG=np.random.RandomState(1337),
+    rng=np.random.default_rng(1337),
```
sappelhoff, line 80: we should move to `random.default_rng` from `random.RandomState` everywhere piece by piece

PR #156, opened 2024-08-29
Reference interpolates instead of deletes channels manually marked as bad, closes #146

### LINKED ISSUE #146: add channels in raw.info['bads'] to unusable channels in Reference class
Currently, PREP assumes you have no prior information about which channels are bad. This isn't necessarily a bad design choice if it's philosophically important that all bad channel identification be automatic, but currently it will result in an error if you have any channels marked as bad in `raw.info['bads']` before running PREP.

The culprit re: the error is in `Reference.__init__`:
```
self.raw.pick_types(eeg=True, eog=False, meg=False)
```
`Raw.pick_types` will not include bad channels in its output by default, and since `Reference` inherits `self.prep_params["ref_chs"] = self.ch_names_eeg` from the `PrepPipeline` class that initializes it (where channels manually marked as bad _haven't_ been removed yet), then the `self.raw.get_data(picks=self.reference_channels)` in `Reference.robust_reference` will throw a missing value error.

So a quick solution would be to change the offending line to 
```
self.raw.pick_types(eeg=True, eog=False, meg=False, exclude=[])
```
so that manually marked bad channels are not unintentionally removed. But they should probably also be added to `Reference.unusable_channels`, so users have the ability to manually tell PREP to leave out certain channels (e.g. if they know the channel is broken in their hardware). 

In the meantime, users can circumvent the problem by just setting `raw.info['bads'] = []` before running PREP. Hopefully that saves someone some trouble :)

john-veillette:
# PR Description

 

Fixes #146, an error that results from the `raw.pick_types` call inside `pyprep.Reference` deleting channels which have been marked as bad prior to PREP. These channels should now be interpolated instead.

A new field `'bad_manual'` is added to `noisy_channels_before_interpolation` dict so there's still a record the channel was interpolated. This happens inside `Reference`; the `NoisyChannels` class is untouched. 

# Merge Checklist

- [ ] (if applicable): the changes are documented in the changelog [changelog.rst][changelog]
- [ ] (if applicable): new contributors have added themselves to the authors list in the [`CITATION.cff`][citationcff] file

[changelog]: https://github.com/sappelhoff/pyprep/blob/main/docs/changelog.rst
[citationcff]: https://github.com/sappelhoff/pyprep/blob/main/CITATION.cff

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -54,6 +54,7 @@ authors:
     - given-names: John
       family-names: Veillette
       affiliation: 'Department of Psychology, University of Chicago, Chicago, IL, USA'
+      orcid: 'https://orcid.org/0000-0002-0332-4372'
 type: software
 repository-code: 'https://github.com/sappelhoff/pyprep'
 license: MIT
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -51,6 +51,9 @@ authors:
     - given-names: Ayush
       family-names: Agarwal
       affiliation: 'Techno India University, India'
+    - given-names: John
+      family-names: Veillette
+      affiliation: 'Department of Psychology, University of Chicago, Chicago, IL, USA'
 type: software
 repository-code: 'https://github.com/sappelhoff/pyprep'
 license: MIT
--- a/docs/authors.rst
+++ b/docs/authors.rst
@@ -10,3 +10,4 @@
 .. _Stefan Appelhoff: https://stefanappelhoff.com/
 .. _Victor Xiang: https://github.com/Nick3151
 .. _Yorguin Mantilla: https://github.com/yjmantilla
+.. _John Veillette: https://github.com/john-veillette
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -31,7 +31,7 @@ Changelog
 
 Bug
 ~~~
-- nothing yet
+- :class:`~pyprep.Reference` now keeps and interpolates channels channels manually marked as bad before PREP, by `John Veillette`_ (:gh:`146`)
 
 Code health
 ~~~~~~~~~~~
--- a/pyprep/find_noisy_channels.py
+++ b/pyprep/find_noisy_channels.py
@@ -163,7 +163,7 @@ class NoisyChannels:
             "bad_by_SNR": self.bad_by_SNR,
             "bad_by_dropout": self.bad_by_dropout,
             "bad_by_ransac": self.bad_by_ransac,
-            "bad_by_manual": self.bad_by_manual
+            "bad_by_manual": self.bad_by_manual,
         }
 
         all_bads = set()
--- a/pyprep/find_noisy_channels.py
+++ b/pyprep/find_noisy_channels.py
@@ -59,6 +59,7 @@ class NoisyChannels:
 
         raw.load_data()
         self.raw_mne = raw.copy()
+        self.bad_by_manual = raw.info["bads"]
         self.raw_mne.pick_types(eeg=True)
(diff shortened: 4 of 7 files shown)


sappelhoff (commented):
Thanks! Could you please make a small modification to an existing test to also cover this behavior?

Also, when would `"bad_manual": ...,` now be printed, and when would it not be printed?

For example, it won't come up in https://github.com/sappelhoff/pyprep/blob/48b73494db5b5de8a6e708278aaef6303cf50b60/pyprep/find_noisy_channels.py#L17, right?

john-veillette:
Sure, I can take a look at the tests, probably on Monday.

Re: `"bad_manual"`, I think it only prints when `logger.info(f"Bad channels: {self.noisy_channels_original}")` is called in `Reference.robust_reference`. So when you run the full PREP pipeline, it won't show up with the [first](https://github.com/sappelhoff/pyprep/blob/48b73494db5b5de8a6e708278aaef6303cf50b60/pyprep/prep_pipeline.py#L187) `NoisyChannels` instance but will during re-referencing steps. (Not sure what that `noisy_detector = NoisyChannels` is actually doing in `prep_pipeline.py`, come to think of it, it seems like the instance is never referenced again.)

Internally, `NoisyChannels` ends up ignoring manually marked bad channels due to the `mne.Raw.pick_types` call on [line 62](https://github.com/sappelhoff/pyprep/blob/48b73494db5b5de8a6e708278aaef6303cf50b60/pyprep/find_noisy_channels.py#L62), so nothing needed to be changed there.

sappelhoff:
> Not sure what that noisy_detector = NoisyChannels is actually doing in prep_pipeline.py, come to think of it, it seems like the instance is never referenced again

Thanks for pointing this out, this is really dead code:

- #157 

> Internally, NoisyChannels ends up ignoring manually marked bad channels due to the mne.Raw.pick_types call on [line 62](https://github.com/sappelhoff/pyprep/blob/48b73494db5b5de8a6e708278aaef6303cf50b60/pyprep/find_noisy_channels.py#L62), so nothing needed to be changed there.

Thanks for looking into this, I just had a look as well and I had forgotten about this. It would be nice to have the same logic in NoisyChannels, that is, report about channels in `raw.info["bads"]` at initialization with a `bad_manual` entry in the dict one can obtain through [`get_bads`](https://pyprep.readthedocs.io/en/latest/generated/pyprep.NoisyChannels.html#pyprep.NoisyChannels.get_bads) ... and handle those channels the same way as those in "bad_by_flat_nan". I think that'd be an improvement to transparency compared to simply dropping them, and the practical outcome is the same.

Furthermore if we implemented this in `NoisyChannels` directly, then robust_reference would also get this information through these lines: https://github.com/sappelhoff/pyprep/blob/48b73494db5b5de8a6e708278aaef6303cf50b60/pyprep/reference.py#L212-L218 and your main goal would also be achieved (being able to pass a raw object with bad channels already marked without them being dropped).

I am sorry that this comment changes the scope of this PR somewhat, but I hadn't looked into the issue deeply enough to understand all implications before you got started.

john-veillette:
Okay, I've added a `"bad_by_manual"` entry to the output of `NoisyChannels.get_bads(as_dict = True)`, and some test coverage for this behavior to `test_find_noisy_channels.py`. (I changed the name from `bad_manual` to `bad_by_manual` since some of the internal logic of `get_bads` depends on the `bad_by_` prefix.) 

Now `"bad_by_manual": ...` should be printed anytime the logger prints bad channels.

sappelhoff:
Thanks a lot @john-veillette!

PR #210, opened 2026-08-10
Accept level names in any case

sappelhoff:
`setup_logging("warning")` raised, because the stdlib only knows upper-case level names:

```python
>>> pyprep.setup_logging("warning")
ValueError: Unknown level: 'warning'
```

`mne.set_log_level("warning")` accepts it, lower case is the spelling people reach for, and it
is what the sibling packages in this stack show in their docs. So the level is upper-cased when
it is a string, and passed through untouched when it is an int.

Tested for `"warning"`, `"WARNING"`, `"Warning"` and `logging.WARNING`.

`pytest tests/test_logging.py`: 57 passed. `ruff check` clean.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -26,7 +26,7 @@ Version 0.9.0 (unreleased)
 
 Changelog
 ~~~~~~~~~
-- Nothing yet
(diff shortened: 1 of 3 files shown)


sappelhoff:
Superseded by #211, which folds this `level.upper()` fix into a new `_as_level` helper as part of making pyprep quiet by default. The changelog entry is reworked there.

